PR #7, opened 2020-06-17
feat: determine library type (single vs paired)

uniqueg:
TBA


src/infer_single_paired.py
```diff
@@ -0,0 +1,382 @@
...
+class End_parser:
+
+    """ Infer Single or Paired """
+
+    def __init__(self):
+        print("\nRunning Parser\n")
+
+    def SRA_contents(self, name, filename):
+
+        """ Runs fastq-dump """
```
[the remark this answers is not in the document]
rohank63: In this function I am not using fastq-dump to split into two Fastq files, as I know if I will do that, resulting files will be very large. All I am doing is, after taking input file from the user, and if it's identifier matches to the SRA format, then I am using fastq-dump to generate just at max 8 lines ( 1 whole spot , that too at as a subprocess , so there is no new resulting file), what this will do is :
1. This will first do a normal check , even though if somehow file's identifier matches the SRA format, but the experiment type is not there in the SRA archive , it will throw an error.
2. Since after returning , one whole spot : depending on the number of lines, accordingly I have written the code in the next function to determine whether its single or paired, followed by additional check ( if there are two files provided by the user ), to match every ReadID in both the files.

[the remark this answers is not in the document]
uniqueg: For input files, I would still prefer to stick to FASTQ format _only_ for now. So SRA doesn't come into the picture, just different formats of ID lines _within a FASTQ file_.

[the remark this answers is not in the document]
uniqueg: Whether it will make sense in the future to support SRA files as input will depend on whether we could run the whole package on SRA files or not. It's not totally out of the question, but it's not sth that has a high priority right now and we would prefer to focus on the basic implementation for now before adding features that we don't really need.

src/infer_single_paired.py
```diff
@@ -0,0 +1,262 @@
...
+                return int(split_id[1])
+            elif id_type == 2:
+                split_id = list(map(str, seq_id.split()))
+                if split_id[0] in seq_id_dict:
+                    seq_id_dict[split_id[0]] += 1
+                else:
+                    seq_id_dict[split_id[0]] = 1
+                return int(split_id[1][0])
+
+    def result_from_count(self, count: list):
```
uniqueg, line 157: Better to make use of the enumerator to make this easier to follow. Compare my solution.

src/infer_single_paired.py
```diff
@@ -0,0 +1,262 @@
...
+            return 0
+        if count[0]*count[1] == 0:
+            if count[1] == 0:
+                return 1
+            elif count[0] == 0:
+                return 2
+        else:
+            return 3
+
+    def read_file1(self, seq_id_dict: dict):
```
uniqueg, line 175: There should be only a single function/method for reading files, with the file parameterized. Compare my solution.

src/infer_single_paired.py
```diff
@@ -0,0 +1,262 @@
...
+            else:
+                if len_seq_id == len_match_2:
+                    return 2
+                else:
+                    return 0
+
+    def check_file1(self):
+        """Opening and checking errors for first file by the user.
+
+        Attribute Error:
```
uniqueg, line 102: According to the Google-style docstring convention, this should go in a section "Raises:".

src/infer_single_paired.py
```diff
@@ -0,0 +1,262 @@
...
+
+    def read_line_in_fastq(self, seq_id: str, seq_id_dict: dict):
+        """This function processes for every seq_id in a fastq file.
+
+        Args:
+            seq_id (str): Sequence identifier.
+            seq_id_dict (dict) : Dictionary to maintain frequency
+            of every seq_id.
+
+        Atrributes:
```
uniqueg, line 133: Section "Attributes:" is reserved for actual attributes of a class.

src/infer_single_paired.py
```diff
@@ -0,0 +1,262 @@
...
+
+class EndParser:
+    def __init__(self, file1_name: str, file2_name: str):
+        """Constructor to assign filenames and initialize file pointers.
+
+        Args:
+            file1_name (str) : First file name provided by the user.
+            file2_name (str) : Second file name provide by the user.
+
+        Attributes:
```
uniqueg, line 49: You have 4 attributes in this case, two file names and two file pointers; for the first two of them, you should copy the description from section "Args:".

src/infer_single_paired.py
```diff
@@ -0,0 +1,262 @@
...
+        if count[0]*count[1] == 0:
+            if count[1] == 0:
+                return 1
+            elif count[0] == 0:
+                return 2
+        else:
+            return 3
+
+    def read_file1(self, seq_id_dict: dict):
+        """Function reads file1 and calls other helper functions.
```
uniqueg, line 176: This (and most others) is not a very helpful description.

PR #40, opened 2020-10-21
feat: subset transcripts from specified organism

### LINKED ISSUE #25: Extract RP transcripts for provided organism
The repository contains a file with transcript sequences of highly expressed ribosomal proteins for hundreds of organisms (prepared for the purpose of inferring the organism from the data, a functionality which has not yet been merged). You can find it at  [`data/transcripts.fasta`](https://github.com/zavolanlab/htsinfer/blob/dev/data/transcripts.fasta). In order to map the reads against the transcriptome in #26 and #27 of the read orientation inference process. we will need to select from this file only the FASTA entries of those transcripts that originate from the organism that was provided through #23 & #24.

To do so, implement a function that parses the identifier lines of [`data/transcripts.fasta`](https://github.com/zavolanlab/htsinfer/blob/dev/data/transcripts.fasta) and selects those entries (identifier and corresponding sequence) that match the 4th field if the provided organism is a string (organism short name) _or_ the 5th column if the provided organism is an integer (taxon ID). See #24 for more info on this. Write the resulting records to a file in the temporary directory created in #33 and return the path to the resulting transcripts file. Make sure you call your new function in the main function `infer()` in the [`htsinfer/infer_read_orientation.py`](https://github.com/zavolanlab/htsinfer/blob/dev/htsinfer/infer_read_orientation.py) module and assign the result to a variable that can be used subsequently to create an index of the file (#26).

danielbatora:
### Description

* extract transcripts of the specified organism from FASTA file
* write results to temporary FASTA file
* for transcripts to be extracted, sequence identifiers in input FASTA file need to adhere to the following format:'
  * needs to consist of at least 4 fragments, separated by the `|` character
  * column 4: organism short name (e.g., `hsapiens`)
  * column 5: taxon identifier (e.g., `9606`)
  * provided organism is compared against _either_ column, if available
  * transcripts are selected only in case of an exact match and if columns are available and any FASTA records can be parsed; otherwise, records are skipped

Resolves #25 

### Type of change

### Checklist

_or do not apply_.

      warnings
      works
      by the proposed changes

together.


issue_25/get_seq.py
```diff
@@ -0,0 +1,51 @@
+# -*- coding: utf-8 -*-
```
uniqueg: Please include your function in the module `htsinfer/infer_read_orientation.py` and call the function in its `infer()` function. There should be an example in there already on how to do that.
You don't need to add author info as this will be tracked by Git / GitHub automatically, on a per-line basis :)

issue_25/get_seq.py
```diff
@@ -0,0 +1,51 @@
...
+"""
+Created on Wed Oct 14 11:25:08 2020
+
+@author: Bátora Dániel
+"""
+
+import os
```
uniqueg: There's some guideline on ordering imports:

1. Group all built-in packages (os, typing)
2. Empty line
3. Group all external packages (Bio)
4. Empty line
5. Group all module imports from _this_ package (does not apply here)

Within each group it's useful (though not prescribed) to order imports alphabetically

issue_25/get_seq.py
```diff
@@ -0,0 +1,51 @@
...
+@author: Bátora Dániel
+"""
+
+import os 
+from Bio import SeqIO
+from typing import Union
+
+
+
+def get_seq(filepath:str,organism:Union[str, int],temp_dir:str) -> str:
```
uniqueg: better: `get_transcripts()` or `get_transcript_seqs()` as these are more expressive

issue_25/get_seq.py
```diff
@@ -0,0 +1,51 @@
...
+    
+    """
+    Selects DNA sequence of the desired organims,
```
uniqueg: Be clear that what we are extracting is transcript sequences. Also, it might be good to include an example in this docstring as we are expecting a very specific layout of the FASTA ID line (4th "column" and 5th "column"), people who do somehow not have access to the original FASTA file will not be able to use the function easily

issue_25/get_seq.py
```diff
@@ -0,0 +1,51 @@
...
+    if not os.path.isdir(temp_dir): 
+        os.mkdir(temp_dir)
+    output_path = os.path.join(temp_dir, organism + ".fasta")
+    reads = []
+    with open(filepath,  "r", encoding = "utf-8") as handle:
+        for record in SeqIO.parse(handle, "fasta"):
+            reads.append(record)
+    
+    hits = []
+    for record in reads:
```
uniqueg: You are actually going through each record _twice_! Why not check whether you want to keep the record while you are looping over them in line 37?

uniqueg: In fact, there is no need to even have a `hits` list because you can just write them out when you loop over them...

issue_25/get_seq.py
```diff
@@ -0,0 +1,51 @@
...
+        if (isinstance(organism, str) and record.description.split("|")[3] == organism) or (isinstance(organism, int) and record.description.split("|")[4] == str(organism)):
+                hits.append(record)
+    SeqIO.write(hits, output_path, "fasta")
```
uniqueg: Instead of checking whether the temporary directory is there, you can just assume it's there. If it's NOT there, the code will throw an OSError (you should add this under section `Raises:` in the docstring). We want to separate concerns in different functions. If we check everywhere everything every function will get bloated and we will have a lot of repeat code. If the person using the function does not use it correctly an error is thrown, that's fair enough. So no need to try to correct their errors, we can take care of that upstream when the function is called.

issue_25/get_seq.py
```diff
@@ -0,0 +1,51 @@
...
+    
+    return output_path
+        
+hits = get_seq(os.path.join(os.getcwd(), "transcripts.fasta"),  "avaga", "C:/github/htsinfer/data/temp_dir")
```
uniqueg: This should go in the `infer()` function of the `htsinfer/infer_read_orientation.py` module

uniqueg: Also, it's better to be explicit (makes it easier to read), so mention the argument names when calling a function, at least if it's more than one arg you are passing.

uniqueg (changes requested):
In addition to the inline changes, the following is required before merging:
- add unit tests for your function
- fix all linter issues in both the function and the unit tests

issue_25/get_seq.py
danielbatora: Can you give me an example on how to do this? Because as it loops over the records, it will only write out the current one and overwrites the previous file. If I store the records in a list, the write function automatically makes the fasta from the list.

mzavolan: Check out https://biopython.org/docs/1.75/api/Bio.SeqIO.html
"If you give a filename, then each time you call write() the existing file will be overwritten. For sequential files formats (e.g. fasta, genbank) each “record block” holds a single sequence. For these files it would probably be safe to call write() multiple times by re-using the same handle."

uniqueg (approved):
Added tests and refactored quite heavily. Please have a look. If you want to know details about specific changes, please ask in this thread.

Thanks for your contribution :-)

uniqueg:
Szia @rockdeme: I see what you did with the `fasz.fasta` ;-)

rockdeme:
> Szia @rockdeme: I see what you did with the `fasz.fasta` ;-)

Actually the idea came from a large collaborative effort by the two of us with @danielbatora :D but how do you know this?

uniqueg:
Az ex-barátnőm magyar volt. Probably wrong, but guess you get what I'm trying to say

rockdeme:
No, the sentence is perfect! Looks like we need to be on a lookout from now on. :D

PR #43, opened 2020-10-28
Issue13 count motifs

### LINKED ISSUE #13: Count motifs in a set of sequences
Given a set of sequences, calculate the number of occurrences of all motifs within a range of length that occur in the input sequences. Return the results as a dictionary.

stinea98:
### Description
This function calculates all the occuring motifs in a set of sequences in a specified length and returns them as a dict.

Fixes # 13

### Type of change

      to not work as expected)
- [ ] Documentation update

### Checklist

_or do not apply_.

      warnings
      works
      by the proposed changes


stinea98:
this is from Lukas Heuberger and Stine Anzböck.

.idea/.gitignore
```diff
@@ -0,0 +1,8 @@
+# Default ignored files
```
uniqueg, line 1: Please add `.idea` to the `.gitignore` file in the project root as we don't want to version-control your personal editor files. You will also need to explicitly remove the files that were already added (all files in `.idea` folder). Find more info online, e.g., here: https://docs.github.com/en/free-pro-team@latest/github/using-git/ignoring-files

This comment applies to the next couple of files, too.

htsinfer/infer_count_motifs.py
```diff
@@ -0,0 +1,91 @@
+"""Infer motif frequency from sample data."""
+
+
+from typing import Dict
+
+
+def count_motifs(input_sequences: list, min_motif_length: int,
```
uniqueg, line 24: Using the builtin `list` and `dict` types for type hinting is only really supported since Python 3.9 (which came out in October). For backwards compatibility it is better to use `List` and `Dict` which you need to explicitly import from the `typing` module.

htsinfer/infer_count_motifs.py
```diff
@@ -0,0 +1,91 @@
...
+                 max_motif_length: int, nucleic_acid: str) -> dict:
+    """Function that calculates the occurrence of all motifs in one or
+    multiple sequences and returns a dictionary with all motifs within
+    the specified length and their occurrence.
+
+        Args:
+            input_sequences (list): list of sequences
```
uniqueg: Since you are using type hints, you can (and probably should) omit the type info from the docstrings, as documentation tooling can automatically parse the type hints and add them to API docs. The problem about adding info to comments rather than in the code is that over time, comments and code tend to diverge, because people forget to update the comments. It's always better to have a single source of truth, and the code is the only really source of truth...

htsinfer/infer_count_motifs.py
```diff
@@ -0,0 +1,91 @@
...
+            min_motif_length (int): minimal length of motif
+            max_motif_length (int): maximal length of motif
+            nucleic_acid (string): type of nucleic acid (dna, rna)
+
+        Returns:
+            sorted_dict (dict): paired motif frequency {motif: frequency }
+
+        Raises:
+            TypeError: input_sequences is not a list
```
uniqueg: If you put named entities like input parameters in docstrings, it is good practice to put them between backticks so that they are rendered and possibly even linked by autodocumentation tooling. So better to write

```text
TypeError: `input_sequences` is not a list
```

etc.

htsinfer/infer_count_motifs.py
```diff
@@ -0,0 +1,91 @@
...
+        raise ValueError('nucleic_acid is not in specified range (dna/rna)!')
+
+    if max_motif_length > len(min(input_sequences)):
+        raise ValueError('max_motif_length is longer than the shortest list')
+
+    motif_freq: Dict[str, int] = {}  # Create empty dictionary (str -> int)
+
+    for sequence_no in range(0, len(input_sequences)):
+        seq = input_sequences[sequence_no]
+        user_choice = "0"
```
uniqueg: If it can be at all avoided, it's not the best idea to make your script interactive because you cannot use it in pipelines anymore and even if run standalone, it's a bit annoying to start a script, then the script runs, possibly for x minutes, then the script stops and waits for user input. It's just not the best user experience. It is better to add a parameter `non-nucleotide characters` to the function and accept a string value that could be either `ignore_chars`, `ignore_seqs` and `include`, then set a default for that parameter (probably `ignore_seqs`).

htsinfer/infer_count_motifs.py
```diff
@@ -0,0 +1,91 @@
...
+                 max_motif_length: int, nucleic_acid: str) -> dict:
+    """Function that calculates the occurrence of all motifs in one or
+    multiple sequences and returns a dictionary with all motifs within
+    the specified length and their occurrence.
+
+        Args:
+            input_sequences (list): list of sequences
+            min_motif_length (int): minimal length of motif
+            max_motif_length (int): maximal length of motif
+            nucleic_acid (string): type of nucleic acid (dna, rna)
```
uniqueg: Indicate with single or double quotes that 'dna' and 'rna' are actually the expected string arguments for this parameter

htsinfer/infer_count_motifs.py
```diff
@@ -0,0 +1,91 @@
...
+            sorted_dict (dict): paired motif frequency {motif: frequency }
+
+        Raises:
+            TypeError: input_sequences is not a list
+            TypeError: min_motif_length or max_motif_length is not an integer
+            ValueError: nucleic_acid is not in specified range
+            ValueError: max_motif_length is longer than the shortest list
+    """
+
+    if not isinstance(input_sequences, list):
```
uniqueg, line 50: The input you'll get in this function will be passed from another function, and the type hints will make sure that we will realize if what we pass is not of the correct type. Thus, when using type hints, it is really only necessary to validate actual raw user input.

Also note that you might feel that, having done your type checks, you are going to be safe when actually using the variables, but in Python, the type of a variable can change between one value and another. Even if you're sure that you don't put any code in between your validation and the place(s) where you use a variable that may change the type, someone else _may_ do so in the future, and the validation might give a false sense of security. If you don't want to have your script just fail in a case like that, the best thing is to put your potentially failing code in a `try - except` block and specifically catch any exceptions that you can think of.

htsinfer/infer_count_motifs.py
uniqueg: Useful to set a default, preferentially `dna`. Always try to make your functions easy to use

htsinfer/infer_count_motifs.py
```diff
@@ -0,0 +1,91 @@
+"""Infer motif frequency from sample data."""
+
+
+from typing import Dict
+
+
+def count_motifs(input_sequences: list, min_motif_length: int,
+                 max_motif_length: int, nucleic_acid: str) -> dict:
+    """Function that calculates the occurrence of all motifs in one or
```
uniqueg: Best to keep these short, precise and free of redundancy. The function signature and context make it clear that this is a function. And the `Returns` section specifies in more detail what is returned.

Better would be something like: "Count motifs on one or more nucleotide sequences"

htsinfer/infer_count_motifs.py
```diff
@@ -0,0 +1,91 @@
...
+        raise TypeError('min_motif_length is not an integer!')
+
+    if not isinstance(max_motif_length, int):
+        raise TypeError('max_motif_length is not an integer!')
+
+    if nucleic_acid.lower() not in ['rna', 'dna']:
+        raise ValueError('nucleic_acid is not in specified range (dna/rna)!')
+
+    if max_motif_length > len(min(input_sequences)):
+        raise ValueError('max_motif_length is longer than the shortest list')
```
uniqueg: Didn't you want to write:
> is longer than the shortest ~list~ _sequence_
?

Still, I don't really get why this is a problem. Shouldn't you make sure the _minimal_ motif length is not bigger than the shortest sequence? I think it's fine if the maximum motif length is never found

htsinfer/infer_count_motifs.py
```diff
@@ -0,0 +1,91 @@
...
+            ValueError: max_motif_length is longer than the shortest list
+    """
+
+    if not isinstance(input_sequences, list):
+        raise TypeError('input_sequences is not a list of sequences!')
+
+    if not input_sequences:
+        raise TypeError('input_sequences is not a list of sequences!')
+
+    if not isinstance(min_motif_length, int):
```
uniqueg: Rather than checking the type I'd rather check if it is a _positive_ integer and that `max` isn't smaller than `min`

htsinfer/infer_count_motifs.py
```diff
@@ -0,0 +1,91 @@
...
+                motif = seq[char: char + motif_length]
+                if all(char in valid_bases for char in motif):
+                    if motif in motif_freq:
+                        motif_freq[motif] += 1
+                    else:
+                        motif_freq[motif] = 1
+                else:
+                    print('motif not valid: ', motif)
+
+    sorted_dict = dict(
```
uniqueg: I wouldn't return a sorted dict, because it takes time to sort and sorting isn't always desired. Therefore it's probably better to let the _consumer_ of the dictionary sort it, because they know better if sorting is needed or not.

tests/test_count_motifs.py
```diff
@@ -0,0 +1,104 @@
+"""Unit tests for infer_single_paired.py"""
```
uniqueg, line 1: The tests are really nice! Well done!  :)

(though probably need some changes after addressing the comments above, sorry about that)

uniqueg (changes requested):

uniqueg:
@stinea98 & @lukasheuberger: sorry, forgot to review your PR earlier. It's really nice and thorough work! Still added quite a few comments for improvement. Would be great if you could address these

lukasheuberger:
@uniqueg thank you for your feedback. we tried to address these points and improved our code. however, it still fails to auto-merge because of the error "R0912: Too many branches (20/12) (too-many-branches)". I assume it is because we have too many if-statements for the error-handling. should we try to get rid of more of these or is it still ok if we get this error?

uniqueg:
@lukasheuberger Yes, please have a try. If you don't manage, you can check online on how to disable specific Pylint warnings, either per line or by adding the error code to the [Pylint config file](https://github.com/zavolanlab/htsinfer/blob/dev/pylint.cfg). Goes for the other issue as well.

htsinfer/infer_count_motifs.py
```diff
@@ -0,0 +1,98 @@
...
+    except KeyError:
+        raise ValueError(
+            'nucleic_acid is not in specified range (dna/rna)!')
+
+    for sequence_no, seq in enumerate(input_sequences):
+        # non_nucleotide_characters == 'ignore_seqs' skips this sequence
+        # and continues with the next sequence
+        if not all(base in valid_bases for base in seq) \
+                and non_nucleotide_characters == 'ignore_seqs':
+            print('sequence', sequence_no+1, 'ignored')
```
uniqueg, line 84: Use `print()` only to debug, use `logging` library for log messages

htsinfer/infer_count_motifs.py
```diff
@@ -0,0 +1,98 @@
...
+            continue
+
+        # collect all motifs and add to dict
+        for motif_length in range(min_motif_length, max_motif_length + 1):
+            for char in range(0, len(seq) - motif_length + 1):
+                motif = seq[char: char + motif_length]
+
+                if all(char in valid_bases for char in motif) or \
+                        non_nucleotide_characters == 'include':
+                    motif_freq = append_motif(motif, motif_freq)
```
mzavolan, line 94: The dictionary of motif counts should not be rewritten again and again. Only the counts of motifs should be updated. One can use defaultdict to take care of initialization.

uniqueg:
Looks pretty good. Resolve conflicts, review and resolve open comments, make sure all tests are reasonable and merge.

uniqueg:
Using PR #55 instead

PR #44, opened 2020-10-28
feat: create/delete temporary directory

### LINKED ISSUE #33: Create temporary directory
Several functions that need to be added to implement the read orientation inference functionality will need to write temporary data to the disk. In order to keep things organized and allow cleaning up temporary data after we are done with the inference, a temporary directory should be created to keep all of the data that is written in subsequent steps.

Implement a function that accepts a directory path, creates a temporary directory in that directory and return the path to the temporary directory. Make sure you call your new function in the main function `infer()` in the `htsinfer/infer_read_orientation.py` module and assign the result to a variable that can be used subsequently (#25, #26, #27, #34). Also make sure to write at the very end of the `infer()` function the code required to delete the temporary directory! All other required calls/code to implement the read orientation inference functionality will go _between_ the creation and deletion of the temporary directory.

yeastethanol:
### Description

* create and delete temporary directory in read orientation inference module

Fixes #33 

### Type of change

### Checklist

_or do not apply_.

      warnings
      works
      by the proposed changes

together.

I do not know how to do the rest.

### DIFF
--- a/htsinfer/infer_read_orientation.py
+++ b/htsinfer/infer_read_orientation.py
@@ -1,7 +1,7 @@
 """Infer read orientation from sample data."""
 
 import logging
-from pathlib import Path
+import os
 import shutil
 from tempfile import mkdtemp
 from typing import Union
@@ -31,7 +31,7 @@ def infer(
 
     # create temporary directory in "../data" directory
     try:
-        tmp_dir = mkdtemp(dir=Path(__file__).resolve().parent.parent / "data")
+        tmp_dir = mkdtemp(dir=os.getcwd())
     except OSError as exc:
         raise OSError("Creation of temporary directory failed") from exc
     LOGGER.info(f"Created temporary directory '{tmp_dir}'")
--- a/htsinfer/infer_read_orientation.py
+++ b/htsinfer/infer_read_orientation.py
@@ -1,24 +1,27 @@
-"""Infer read orientation from sample data.""" def infer(iputDir):
+"""Infer read orientation from sample data."""
+
+
+def infer(iputDir):
     """Main function coordinating the execution of all other functions.
     Should be imported/called from main app and return results to it.
     """
-    # implement me
     
-    #create .tmp
-    mkTmpDir(inputDir)
+    #create .tmp directory in "inputDir" and retuns .tmp path
+    tmpDir = mkTmpDir(inputDir)
     
-    #Do something then change directory back to previous directory to allow .tmp to 
-    #be deleted
+    #Call functions of issues #25, 26, 27 and 34
     
-    os.chdir(inputDir)
-    print("Changed working directory to %s " % tmpPath)
     
     # delete .tmp
-    rmTmpDir()
+    rmTmpDir(tmpDir)
     
     
 def mkTmpDir(inputDir):
+    """Function to create a temporary directory for the storage of 
+    temporary data."""
+  
   import os
+
   # define the name of the directory to be created
   tmpPath = inputDir + "/.tmp"
   
@@ -29,16 +32,17 @@ def mkTmpDir(inputDir):
   else:
       print ("Successfully created the directory %s " % tmpPath)
   
-  os.chdir()
-  print("Changed working directory to %s " % tmpPath)
+  return (tmpPath)
+      
       
 def rmTmpDir(tmpPath):
-  import os
+    """Function to remove the temporary directory."""
   
+  import os
   #remove temporary directory
   try:
-    os.rmdir(tmpPath) 
-except OSError:
+    os.rmdir(tmpPath)
+  except OSError:
     print ("Deletion of the directory %s failed" % tmpPath)
   else:
     print ("Successfully deleted the directory %s" % tmpPath)
--- a/htsinfer/infer_read_orientation.py
+++ b/htsinfer/infer_read_orientation.py
@@ -1,65 +1,39 @@
 """Infer read orientation from sample data."""
 
-from typing import Union
+import logging
+import os
+import shutil
+from tempfile import mkdtemp
 
-def infer(
-    file_1: str,  # pylint: disable=unused-argument
-    file_2: str = None,  # pylint: disable=unused-argument
-    organism: Union[int, str] = "hsapiens",  # pylint: disable=unused-argument
-) -> str:
-    """Infers read orientation for single- or paired-ended sequencing libraries
-    in FASTQ format.
+LOGGER = logging.getLogger(__name__)
 
-    Args:
-        file_1: File path to read/first mate library.
-        file_2: File path to second mate library.
-        organism: Source organism of the sequencing library; either a short
-            name (string, e.g., `hsapiens`) or a taxon identifier (integer,
-            e.g., `9606`).
 
-    Returns:
-        LIBTYPE string according to Salmon documentation, cf.
-        https://salmon.readthedocs.io/en/latest/library_type.html
+def infer():
+    
+    """Creates a temporary directory (tmp_dir) in the input 
+    directory (input_dir). Returns the path of tmp_dir.
+    tmp_dir is used for the storage of temporary data 
+    created when performing the main function. 
+    After completion of main function, tmp_dir is deleted.
     """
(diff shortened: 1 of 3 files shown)


htsinfer/infer_read_orientation.py
```diff
@@ -22,5 +21,45 @@ def infer(
...
+    # call functions of issues #25, 26, 27 and 34
+    
+    # delete .tmp
+    rmTmpDir(tmpDir)
+        
     # implement logic
     return "U"
+    
+
+def mkTmpDir(inputDir):
```
uniqueg: It is a convention to use "snake case" for function names and variables, and "camel case" for classes. So in this case, as you are writing a function, the name of the function should be `mk_tmp_dir(input_dir)` (and similarly for the other function below). See [Python's style guidelines](https://www.python.org/dev/peps/pep-0008/) for more details

htsinfer/infer_read_orientation.py
```diff
@@ -22,5 +21,45 @@ def infer(
...
+    """Function to create a temporary directory for the storage of 
+    temporary data."""
+  
+    import os
```
uniqueg: Please put imports at the top of the module, right below the module's docstring. FYI, the same [Python style guidelines](https://www.python.org/dev/peps/pep-0008/#imports) mentioned above also suggest an order in which imports should be listed, with standard/builtin packages first, followed by third party libraries (the ones we have to install) and then the imports of modules from this very package (i.e., `htsinfer`), and with each group (builtin, third party, current package) being separated by an empty line. For example:

```py
import os  # this is a builtin library, we don't have to install it specifically with PyPI
import sys  # also builtin

import Bio  # BioPython is a third party library, it needs to be specifically installed

from htsinfer import infer_read_orientation  # this is from the current package
```

Specifically, if you move the `import os` to the beginning of the file, you need to group it together with the import from `typing` which is also builtin, and within each group it's good practice to follow alphabetical order of the package name:

```py
import os
from typing import Union
```

htsinfer/infer_read_orientation.py
```diff
@@ -22,5 +21,45 @@ def infer(
...
+    
+    # delete .tmp
+    rmTmpDir(tmpDir)
+        
     # implement logic
     return "U"
+    
+
+def mkTmpDir(inputDir):
+    """Function to create a temporary directory for the storage of
```
uniqueg: Please always make sure to describe input params, the return value and potentially raised errors in the docstring and follow the Google-style convention to do so. More examples for this are in the `infer_single_paired.py` module.

htsinfer/infer_read_orientation.py
uniqueg: Make sure to add type hints to any input params of your functions as well as to the return value. In this case, your signature would become:

```py
def mk_tmp_dir(input_dir: str) -> str:
```

because you are expecting a string input and will return a string output

htsinfer/infer_read_orientation.py
```diff
@@ -22,5 +21,45 @@ def infer(
...
+    
+
+def mkTmpDir(inputDir):
+    """Function to create a temporary directory for the storage of 
+    temporary data."""
+  
+    import os
+
+    # define the name of the directory to be created
+    tmpPath = inputDir + "/.tmp"
```
uniqueg: This is not a very good name for the temporary directory because the chance of collisions is quite high and we don't want to risk overwriting data. There is a standard/builtin module `tempfile` in Python for such cases and it's function `mkdtemp()` is specifically made for our purpose. So you add `from tempfile import mkdtemp` to the top of the file and then create a directory the directory under `input_dir` with `mkdtemp(dir=input_dir)`. This will both generate a name _and_ create the input directory and then return the name of the directory.

In fact you can just simply return it like so:

```py
return mkdtemp(dir=input_dir)
``` 

This is all you need to do in this function. However, since your function will then just be a single line, it doesn't make sense to have a function for this at all, so you can just add this call to the main `infer()` method:

```py
tmp_dir = mkdtemp(dir=__path__)
```

This will create the temporary directory in the directory where the current directory is located.

htsinfer/infer_read_orientation.py
```diff
@@ -22,5 +21,45 @@ def infer(
...
+  
+    try:
```
uniqueg, line 33: Handling errors is better done as much upstream as possible in order to increase reusability of a function and separate concerns. So the `try - except` block should rather go in the calling context (or possibly even further up), not the function itself.

Another issue is that you are not re-raising the error, just printing an error message. This is a problem, because after printing, the program will just continue to execute _as if the directory were created_. But if we don't have a temporary directory, we can't continue with the code. So what you need to do is re-raise the error after logging your message. 
Also, never use `print()` for logging, use a proper (via the [`logging`[(https://docs.python.org/3/library/logging.html) module) instead. See `infer_single_paired.py` for an example. In this case, for the error log message, we can also raise the same (or a different/custom) error with a custom error message _from_ the raised error.

Finally, I highly recommend you to use ["f strings"](https://realpython.com/python-f-strings/#f-strings-a-new-and-improved-way-to-format-strings-in-python) for string interpolation, as they are shorter and much easier to read and write. In this case though, we don't know the name `mkdtemp()` generated, so there's nothing to interpolate. We'd just log that creation of the temporary directory failed.

All put together, your code would become (in `infer()`):

```py
import logging
from tempfile import mkdtemp
from typing import Union

LOGGER = logging.getLogger(__name__)


def infer(...):  # incomplete, of course use the whole function signature, docstring etc.

    # create temporary directory
    try:
        tmp_dir = mkdtemp(dir=__path__)
    except OSError as exc:
        raise OSError("Creation of temporary directory failed") from exc
    LOGGING.info(f"Created temporary directory {tmp_dir}.")
```

htsinfer/infer_read_orientation.py
```diff
@@ -22,5 +21,45 @@ def infer(
...
+    return (tmpPath)
+      
+      
+def rmTmpDir(tmpPath):
+    """Function to remove the temporary directory."""
+  
+    import os
+    #remove temporary directory
+    try:
+        os.rmdir(tmpPath)
```
uniqueg: This only deletes an _empty_ directory. Instead you need to remove the directory and all its contents. See here for how to do this: https://stackoverflow.com/questions/6996603/how-to-delete-a-file-or-folder

Again, you won't need a function for that, just put it at the end of the `infer()` function, just before the `return` statement. Add the `try - except` and log messages as described above.

requirements.txt
```diff
@@ -0,0 +1,8 @@
+coverage==5.0.3
```
uniqueg: We don't want to have this file at all. Instead we are using `environment.yml` to have Conda handle dependencies. Please remove it.

uniqueg (changes requested):
The general approach was good, but there are much easier (and safer) ways to get this done in Python with one-liners, so you won't need special functions for creating and removing the temp directory at all. I have added quite a lot of thoughts on most of your code just so that you know what can be improved, maybe it will be helpful to you even if most of it you won't need for this specific issue.

As your code now will end up only in the main function, you don't need to write unit tests for the creation/deletion functions. However, you would need to write additional tests for `infer()` that actually test its successful and unsuccessful behavior. For this you need to find, or rather set up, cases where (1) creation and (2) deletion of the temporary directory fails. This is going to be the trickiest part. The simplest way to do this is probably by "monkeypatching" the result of the `mktemp()` and `rmtree()` functions. See here on how to "monkeypatch" external functions with `pytest`: https://docs.pytest.org/en/stable/monkeypatch.html

uniqueg:
I've added the tests. Please have a look here: https://github.com/zavolanlab/htsinfer/pull/44/files

uniqueg:
Otherwise, the implementation of the logic to create and delete the temporary directory was fine, but the rest of the `infer()` function should have been left alone. It's supposed to be the main function of this module, which coordinates the calling of all other functions to get the result, so it needs to have the input files as arguments, return a result string and so on. I've re-added it and it should be fine now. You can have a look at the changes, I think they should be fairly self-explanatory...

Thanks a lot for your contribution :)

uniqueg (approved):

PR #45, opened 2020-11-01
implemented motif search

### LINKED ISSUE #35: Find motif matches within a read
For constructing the position-dependent profile of motif occurrences and counting the reads that match individual motifs (issues #18 and #19), we need a functionality of finding where a specific motif occurs within a read. We could do this by sliding the motif from left to right, starting from a minimum overlap between motif end and read start and ending at some minimum overlap between motif start and read end, reporting all of the consistent matches as a list of tuples, (position:fraction of motif matched). Running this function over all reads and all motifs will be time-consuming, so minimize the number of tests that are done to decide whether the motif matches or not.

ymoon06:
### Description

takes motif, read, min_overlap and optional parameter and performs  a motif search in the read. 
It returns a list of tuple (start of overlap in the read, fraction of overlap). uses list comprehension for fast execution. 
Used == for fast string comparison.
test files are also designed to test all possible corner cases.

Fixes # (issue)

### Type of change

I am not sure what this part is about. But I can say I don't have bugs in the unit test file.(I fixed the bug along I code bits by bits)

- [ ] New feature (non-breaking change which adds functionality)
      to not work as expected)
- [ ] Documentation update

### Checklist

_or do not apply_.

      warnings
      works
      by the proposed changes

together.


ymoon06:
fixed my style so that it passes flake8 tests. Seems like passed successfully. 
tested functions with all corner cases I can think of. 
1 more test to do is to test with a lot of reads and motifs so that we can examine how it performs under stressful condition.
Thank you.

read_motif_v6.py
```diff
@@ -0,0 +1,42 @@
+# -*- coding: utf-8 -*-
+"""
+Created on Mon Nov  2 01:28:58 2020
+
+@author: YB Moon (ymoon06)
+"""
+# assume motif <read
```
mzavolan: Provide a description of the input-output behavior of the function and of all parameters.

read_motif_v6.py
```diff
@@ -0,0 +1,42 @@
...
+# use_N is for further implement as a parameter
```
mzavolan: What is use_N?

read_motif_v6.py
```diff
@@ -0,0 +1,42 @@
...
+def find_overlaps(motif, read, min_overlap, use_n=False):
+    #check the type of arguments
+    if not isinstance(motif, str) or\
+        not isinstance(read, str) or\
+        not isinstance(min_overlap, int) or\
+        not isinstance(use_n, bool):
+        raise TypeError('pass the right argument')
```
mzavolan: Message should be more precise, e.g. 'Incorrect argument type: [argument]'

read_motif_v6.py
```diff
@@ -0,0 +1,42 @@
...
+    # passing min_overlap <1 does not make sense
+    if len(motif)==0 or min_overlap <1:
+        raise ValueError('pass the right value')
```
mzavolan: Make message more precise

read_motif_v6.py
```diff
@@ -0,0 +1,42 @@
...
+        
+    # compute partial overlaps of the motif at the start of the read (1st case)
+    #list comprehension is faster. parallelizing is allowed
+    partial_overlaps_start=[]
+    if len(read)>= min_overlap:
+        #doesnt make sense to do len(motif)+1 because then it would be the 2nd case
```
mzavolan: Please try to keep the comments concise. They should be readable to someone that may want to pick up the code and develop it further.

read_motif_v6.py
```diff
@@ -0,0 +1,42 @@
...
+        #you start from min_overlap because over lap should satistfy certain threshold
+        #As in motif[len(motif)-ov:], you dont need to specify the end, python understands this is until end 
+        partial_overlaps_start= [ (0, ov/len(motif)) for ov in range(min_overlap, len(motif)) if read[0:ov]==motif[len(motif)-ov:]]
+    
+    # compute matches(overlaps) of the motif inside the read (2nd case)
+    # you compare the full motif length
+    full_overlaps=[]
+    if len(read)>-len(motif):
```
mzavolan: Is this test necessary? lengths will be >= 0

read_motif_v6.py
```diff
@@ -0,0 +1,42 @@
+# -*- coding: utf-8 -*-
+"""
+Created on Mon Nov  2 01:28:58 2020
+
+@author: YB Moon (ymoon06)
+"""
+# assume motif <read
+# use_N is for further implement as a parameter
+def find_overlaps(motif, read, min_overlap, use_n=False):
```
mzavolan: Can you please add an option to report only matches that are fully contained in the read?

test_motif_shortened.py
```diff
@@ -0,0 +1,83 @@
...
+# further to implement. to give stress and see if it works well. (didnt have time to finish this)
+class TestLongReads(unittest.TestCase):
+
+  def test_file_1(self):
+    pass
+
+
+if __name__ == '__main__':
+  unittest.main()
+  #print('success')
```
mzavolan: add new line at the end of the line

test_motif_shortened.py
```diff
@@ -0,0 +1,83 @@
...
+    self.assertTrue(len(rv) == 2)
+    self.assertTrue(rv[0] == (0, 2/3))
+    self.assertTrue(rv[1] == (5, 1))
+
+  def test_multi_match_end(self):
+    rv = find_overlaps(motif = "GTAAA", read = "ACGGTAAAAGT", min_overlap = 2, use_n = False)
+    self.assertTrue(len(rv) == 2)
+    self.assertTrue(rv[0] == (3, 1))
+    self.assertTrue(rv[1] == (9, 2/5))
+
```
mzavolan: Can you implement the situation of no match whatsoever?

read_motif_v6.py
mzavolan: Don't you want to check that the strings are both in upper or lower case?

mzavolan (changes requested):

ymoon06:
I didn't upload test file because it had lots of errors with respect to flake8 style.
Don't know how they are visible because I didnt upload it it.
But I think I fixed everything you requested with respect to the motif find function.
If time allowed, I will fix test file as well and upload it separately.
Thank you.

read_motif_v6.py
```diff
@@ -0,0 +1,73 @@
...
+        raise ValueError('length of read must be bigger than 0')
+    if min_overlap < 1:
+        raise ValueError('min_overlap must be longer than 1')
+    if len(read) < len(motif):
+        raise ValueError('read should be >= than motif')
+    if not motif.isupper():
+        raise ValueError('motif should be all big characters')
+    if not read.isupper():
+        raise ValueError('read should be all big characters')
+    """compute partial overlaps of the motif at the start of the read"""
```
uniqueg: Use `"""` comments only for docstrings, otherwise use `#` comments. Docstrings are the documentation strings that are expected as the first statement of a module, class, function and method.

read_motif_v6.py
```diff
@@ -0,0 +1,73 @@
+# -*- coding: utf-8 -*-
+"""
```
uniqueg: Please only add a _single_ docstring. Also, please omit the author information, this is automatically tracked by Git, on a line-per-line basis. Adding this here will lead to artifacts, because these metadata are never properly updated when maintaining the code

read_motif_v6.py
```diff
@@ -0,0 +1,73 @@
+# -*- coding: utf-8 -*-
```
uniqueg: Please add your code to the `htsinfer/infer_read_layout.py` module rather than its own module

read_motif_v6.py
uniqueg: Please have a look at and follow the style of Google-style docstrings: https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html

read_motif_v6.py
```diff
@@ -0,0 +1,73 @@
...
+"""
+output:
+    returns a list of tuple(a,b) where:
+    a is overlap start position of read
+    b is fraction of motif that overlaps
+"""
+
+
+def find_overlaps(motif, read, min_overlap, full_contain=False):
+    """check the type of arguments """
```
uniqueg: Please add a complete Google-style docstrings: https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html

read_motif_v6.py
```diff
@@ -0,0 +1,73 @@
...
+"""
+"""
+output:
+    returns a list of tuple(a,b) where:
+    a is overlap start position of read
+    b is fraction of motif that overlaps
+"""
+
+
+def find_overlaps(motif, read, min_overlap, full_contain=False):
```
uniqueg: Please add type hints to indicate the data types of the input parameters as well as the return type: https://docs.python.org/3.8/library/typing.html

uniqueg (changes requested):

ymoon06:
fixed everything except one. If I put my file into the directory then the coverage decreases to 84% or so without any reason. if I put it outside then coverage does not decrease. do you have any idea on this? It is the same code and it is because of the error of my code.

ymoon06:
fixed everything, added a separate test file testing all that I have implemented including arguments.
One thing is that I didnt include it inside the existing file because it caused an build error.
May I ask if it is ok?

uniqueg:
It will be fine like that, thank you very much! If @zavolan is still happy with the logic (I think she was), I will make any final changes, if necessary, and merge. Might take a few days though

uniqueg (approved):

mzavolan (approved):
Type checks could be done via mypy/linter.

PR #47, opened 2020-11-02
Create random seqs

### LINKED ISSUE #12: Randomize nucleotide sequences
Given a set of input sequences, generate randomized variants using a markov process. The order of the markov chain to be considered is a parameter. More concretely, we want to generate shuffled variants of the initial sequences that are similar in some way, but are not constructed to contain specific adaptor/barcode motifs. The simplest way to generate such variant sequences is to *shuffle* (permute the letters of) the input sequences. However, mammalian genes/genomes have some very distinct nucleotide composition biases such as being relatively depleted in CG dinucleotides (relative to the expected frequency of the dinucleotide given the frequencies of C's and G's), which, if ignored in the shuffling, will lead to the misestimation of frequencies of these motifs relative to the genome-average. To better preserve genome composition statistics in the shuffling, we would have to enforce that dinucleotide frequencies of the input sequences are maintained. An approximation that is simpler to implement is to maintain dinucleotide frequencies not strictly, but *on average*, which we can do by computing the frequencies of observing A,C,G,T following each of the A,C,G,T nucleotides and then generating as many and as long random sequences as the initial sequences, extending a sequence base by base, and choosing the next base given the current base and the probabilities we just mentioned.

Another parameter could be the number of sets of randomizations to generate.

mischa-m:
### Description

Implemented a function randomize_nucleotide_sequence() as called for by the issue Randomize nucleotide sequences #12.

Fixes #12

### Type of change

### Checklist

_or do not apply_.

      warnings
      works
      by the proposed changes

together.


htsinfer/rand_nuc_seq.py
```diff
@@ -0,0 +1,249 @@
...
+        number_random_seq x randomised sequences of equal length to the input
+        sequence in a list.
+
+    Raises:
+        ValueError: input_string contains too many characters other than ATGCN.
+        ValueError: The percentage of N's one of the input sequences is too high
+    """
+    
+    # Error if any of the input is not a char
+    if not all(isinstance(n, str) for n in input_sequences):
```
uniqueg: It is possible that we will be dealing with millions of input sequences, so you should try to refactor your code in a way that you loop over the sequences only once. Just make this check further down in the third loop

htsinfer/rand_nuc_seq.py
```diff
@@ -0,0 +1,249 @@
+from collections import Counter
+
+import numpy as np
+
+def randomize_nucleotide_sequence(input_sequences: list,
+    number_random_seq: int = 1, min_prob: float = 0) -> list:
+    """
+    Returns randomised sequence adhering to the dinucleotide probabilites.
```
uniqueg: Please clarify what "dinucleotide probabilities" you are referring to (e.g., "found in a set of input sequences").

htsinfer/rand_nuc_seq.py
```diff
@@ -0,0 +1,249 @@
...
+
+    Args:
+        input_sequences: List of input sequences as strings.
+        number_random_seq: Number of randomised sequences returned per input
+            sequence.
+        min_prob: Minimal probability of dinucleotide to appear. If set to -1
```
uniqueg: It's a really nice idea to include this, but I don't find it very intuitive that setting it to 0 will actually mean that 0 is _not_ possible. I would switch around what 0 and -1 mean here...

htsinfer/rand_nuc_seq.py
```diff
@@ -0,0 +1,249 @@
+from collections import Counter
+
+import numpy as np
+
+def randomize_nucleotide_sequence(input_sequences: list,
```
uniqueg: Type hints with literal `list` and `dict` are only really supported since Python 3.9, which is only a couple of weeks old. For backwards compatibility it is better to use `List` and `Dict` here, which you need to import from the `typing` package.

Below the import from `collections`, add:

```py
from typing import list
```

Then replace the type hints for `list` with `List`.

htsinfer/rand_nuc_seq.py
```diff
@@ -0,0 +1,249 @@
+from collections import Counter
```
uniqueg: Please place the code/function in the `htsinfer/infer_read_layout.py` module rather than in its own module

htsinfer/rand_nuc_seq.py
```diff
@@ -0,0 +1,249 @@
...
+    """
+    
+    # Error if any of the input is not a char
+    if not all(isinstance(n, str) for n in input_sequences):
+        raise ValueError(
+            "Not all input sequences are of type 'str'"
+        )
+
+    # Convert input strings to upper case
+    input_seqs_upper = []
```
uniqueg: Again, you go over all sequences here. Please place the code to capitalize seqs in the next loop

htsinfer/rand_nuc_seq.py
```diff
@@ -0,0 +1,249 @@
...
+    
+    for i, sequence in enumerate(input_sequences):
+        # Error if empty sequence is entered
+        if len(sequence) == 0:
+            raise ValueError(
+                f"Sequence {i} is empty"
+            )
+        # Count the occurences of N's
+        count = Counter(sequence)
+        if count["N"] >= 0.05 * len(sequence):
```
uniqueg: It's not good practice to "hardcode" values. Please make `0.05` a parameter to the function (e.g., `n_freq`)

htsinfer/rand_nuc_seq.py
```diff
@@ -0,0 +1,249 @@
...
+            raise ValueError(
+                f"Percentage of N in input sequence  {input_sequences.index(sequence)} is "
+                f"over the limit of 5 percent. Sequence: \n"
```
uniqueg: Parameterize the `5` here with `n_freq`

htsinfer/rand_nuc_seq.py
```diff
@@ -0,0 +1,249 @@
...
+                f"{sequence}"
+        )
+
+        # Check if there are chars other than "ATGCN"
+        for i,char in enumerate(sequence):
+            if char not in "ATGCN":
+                raise ValueError(f"Input sequence "
+                f"{input_sequences.index(sequence)} contains characters other "
+                f"than 'ATGCN'.\n"
+                f"First occurance:\n"
```
uniqueg: Occurence

uniqueg (changes requested):
This is really nicely written. I have added a couple of comments, but don't have time to finish up the review now. I will more comments later. In the meantime you can address the comments that are already in.

htsinfer/rand_nuc_seq.py
mischa-m: I agree. Actually, Thomas and I thought to leave the min_prob out for now. Apparently, we forgot to clean up the docstrings properly, as we were planning to implement it at the beginning. Our current solution is that, when the dinucleotide occurrences are counted to calculate the markov matrix, we start counting at 1 (so we assume that each combination has already occurred once). Thus, if a certain combination never occurs in the input, it still gets a probability. For short input sequences this is not the most elegant solution and leads to incorrect probabilities, but the problem should be negligible if the input sequences are of sufficient length. We did not do it with the min_prob as we were a bit short on time and wanted to submit something before tomorrow, so maybe we can give it a few more minutes and try to implement it properly.

mzavolan: Your solution includes a 'pseudo-count' of 1, which is fine. As you hint in the above explanation, the pseudo-count value sets the scale for how much data you need to see before you start trusting the counts. If you want to implement an additional parameter, it should be this pseudo-count.

htsinfer/infer_read_layout.py
```diff
@@ -1,8 +1,276 @@
...
+        input_seqs_upper.append(sequence)
+
+        # Error if empty sequence is entered
+        if len(sequence) == 0:
+            raise ValueError(
+                f"Sequence {i} is empty"
+            )
+        # Count the occurences of N's
+        count = Counter(sequence)
+        if count["N"] >= n_freq * len(sequence):
```
mzavolan, line 74: This test is too strict. There should at least be a flag that allows the program to continue if some of the sequences in the input file are not valid (because they contain too many N's or they contain characters that are not [A, C, G, T, N]

htsinfer/infer_read_layout.py
```diff
@@ -1,8 +1,276 @@
...
+    for i in range(len(nucl)):  # i represents the row
+        tot = np.sum(result[i])
+        for j in range(len(nucl)):  # j represents the column
+            result[i, j] = result[i, j]/tot
+
+    return result
+
+
+def count_occurences(string, substring):
+    """Count occurence of base"""
```
uniqueg, line 166: Actually counts substrings, not (individual) bases

uniqueg: Docstring incomplete

uniqueg:
This looks pretty good but needs some changes (see comments). Review, implement and complete tests, then merge.

uniqueg:
mypy fails in Travis because type hints are not available for numpy. Ignore this error.

uniqueg:

PR #48, opened 2020-11-03
Issue15

### LINKED ISSUE #15: Compute histograms of inter-motif distances
Given a set of motifs and a set of sequences, compute histograms of inter-motif distances for all pairs of motifs and return pairs for which there is a significantly preferred distance. That is, loop across all sequences and record the distance between that start positions of the two motifs in every sequence. Note the symmetry, namely that distances between motif1 and motif2 will be the negatives of distances between motif2 and motif1. To not deal get twice the same underlying configuration, we can keep track only of non-negative distances. If there really is a preferred distance, this distance should occur more often than all others. We will have to decide on a cutoff. Here we could use the statistics we compute for generating the randomized sequences to calculate how often we expect specific motifs at specific distances. But maybe a simpler rule of thumb would work well enough, e.g. for each motif, the most frequent distance, as long as it's x fold higher than the average or than the second most common distance.

dcpb0:
### Description

Compute histograms of inter-motif distances.
Fixes # (15)

### Type of change

      to not work as expected)
- [ ] Documentation update

### Checklist

_or do not apply_.

      warnings
      works
- [ ] New and existing unit tests pass locally with my changes and I have not
      by the proposed changes

together.

Comment:
We have used the test files to see if the code is working. Also, we have documented in the code. We think it is clear what is doing what. However, we are not sure about the unticked boxes what they exactly mean.


issue15_code.py
```diff
@@ -0,0 +1,111 @@
...
+def motifHistogram(seq1, seq2, directory,nrbins):
+    
+    #-------------------------------------------------------------------------
+    #create motif file of the 2 input motifs forward and revers
+    instances = [Seq(seq1), Seq(seq2)]
+    m = motifs.create(instances)
+    r = m.reverse_complement()
+    #-------------------------------------------------------------------------
+    
+    #get all reads of files in the folder
```
mzavolan, line 28: The script is doing more than specified, the input that was specified was a set of sequences.

issue15_code.py
```diff
@@ -0,0 +1,111 @@
...
+            motif2r=[]
+        
+            # saves the positions in the forward motifs
+            for pos, seq in m.instances.search(sequence): 
+                if seq == m.instances[0]: #saves the positions of the first motif
+                    motif1f.append(pos)
+                if seq == m.instances[1]: #saves the positions of the second motif
+                    motif2f.append(pos)
+                    
+            # saves the positions of the reverse motifs
```
mzavolan, line 75: Reverse complement is not needed.

issue15_code.py
```diff
@@ -0,0 +1,111 @@
...
+                    motif2f.append(pos)
+                    
+            # saves the positions of the reverse motifs 
+            for pos, seq in r.instances.search(sequence): 
+                if seq == r.instances[0]: #saves the positions of the first motif
+                    motif1r.append(pos)
+                if seq == r.instances[1]: #saves the positions of the second motif
+                    motif2r.append(pos)
+     
+            # calculate  distance between forward motif1 and motif2, and save it in distance array
```
mzavolan, line 82: Positive and negative distances are not equivalent. The specification of the issue describes how they should be handled.

uniqueg:
Not sure how good substance is (doing quite a bit of extra work, may include some faulty code), but left open as it still might be of use. Check carefully to make sure code adheres to spec defined in #15.

uniqueg:

PR #50, opened 2020-11-06
Paul Ritsche: Issue # 14 Calculate motif enrichment 

### LINKED ISSUE #14: Calculate motif enrichment
Given two dictionaries of motif counts (foreground and background), calculate the enrichment (ratio of the probabilities of occurrence of the motif in foreground and background) and p-value (assuming binomial sampling with average frequency taken from the background set) for all motifs in the foreground set. Return these as a dictionary.

PaulRitsche:
### Description

The changes in this pull request address the issue 14. The new function included
in htsinfer is called motif_enrichment. The function calculates the enrichment of 
motifs of similar length in a given foreground and background dictionary. 
Additionally, the function calculates the respective p-value for motifs in the 
foreground dictionary based on a binomial distribution using the average 
probability of occurrence in the background dictionary. Finally, the function
returns a dictionary containing the motifs of the foreground dictionaries as keys
and their respective enrichment and p-value as values. 

Fixes # 14

### Type of change

### Checklist

_or do not apply_.

      warnings
      works
      by the proposed changes


PaulRitsche:
@uniqueg 
I wasn't quite sure how to add you and mihaela as reviewers, I hope that's not too bad.

uniqueg:
@PaulRitsche Thanks a lot. I think you can't assign us, but it's fine. We'll look at it when we get to it, will be a few days

htsinfer/motif_enrichment.py
```diff
@@ -0,0 +1,100 @@
...
+    Returns:
+        Dictionary including motifs, sum of counts of motifs with
+        similar lenght, enrichment score and p-value
+
+    Example:
+        foreground = {"TGATTC": 5, "TAAACC": 3, "AAGTTACCT": 1,
+        "AAGCCTT": 1, "AGTTCTA": 1, "TTTCCCG": 5}
+        background = {"TGATTC": 3, "TAAACC": 5, "AAGCCTTAT": 1,
+        "AGTTCTA": 1, "TTTCCCG": 5, "TTGGAA": 7}
+        >>> motif_enrichment(foreground, background)
```
mzavolan, line 26: The return value in the example is not consistent with the description above (there is no sum of counts returned).

htsinfer/motif_enrichment.py
```diff
@@ -0,0 +1,100 @@
...
+    # Calulates sum of motifs of same length
+    for i in list(foreground.keys()):
+
+        if len(i) in length_list_f:
+            new_dict_foreground[len(i)] += foreground[i]
+        else:
+            new_dict_foreground[len(i)] = foreground[i]
+            length_list_f.append(len(i))
+
+    # Calculates probability of occurrence of motifs
```
mzavolan: This calculates the frequency of occurrence of motifs of a given length among all lengths. Instead, it should calculate the frequency of occurrence of one particular motif among all motifs of that specific length. Using the example at the top, among motifs of length 6 we have in the foreground:
"UGAUUC": 5, "UAAACC": 3
and in the background
"UGAUUC": 3, "UAAACC": 5, "UUGGAA": 7
To account for the possibility that the foreground and background motif sets are not perfectly overlapping, we would create the foreground and background dictionary such that they contain the same motifs, adding a pseudocount to all motifs before calculating their frequencies. In the example above, we would extend the foreground dictionary to also contain UUGGAA, and then, by adding a pseudo-count of 1 to each motif, we would have the following foreground and background dictionaries
"UGAUUC": 6, "UAAACC": 4, "UUGGAA": 1
and in the background
"UGAUUC": 4, "UAAACC": 6, "UUGGAA": 8
Then the frequency of UGAUUC in foreground would be 6/11, in background 4/18, enrichment value would be 6*18/(4*11) and the p value will be calculated based on 6 out of 11 occurrences assuming a probability of 4/18.

PaulRitsche: Thanks for the feedback. 
I'll try and implement it and push the changes once their ready.

PaulRitsche: Given your example, do I need to increase the pseudocode for all motifs of the same length in the dictionary or just of the ones that were included in the respective dictionary from the beginning? Just in case more than one motif of the same length needs to be added to the same dictionary. If we would add another motif, say "UGCUGC" to foreground, would the foreground look like this: "UGAUUC": 7, "UAAACC": 5, "UUGGAA": 2, "UGCUGC": 1 or like this: "UGAUUC": 7, "UAAACC": 5, "UUGGAA": 1, "UGCUGC": 1.

mzavolan: You would need to add the same pseudocount to all motifs

uniqueg:
Looks okay, but needs some cleaning, fixing of tests and possibly some refactoring.

uniqueg:

PR #51, opened 2020-11-08
Count occurrences of adaptors/contaminants

### LINKED ISSUE #19: Count occurrences of adaptors/contaminants
Given a set of known adaptors/contaminants and an input fastq file, count the number of matches to potential adaptors/contaminants. Similar to issue #19, but instead of making position-dependent profiles, the number of matches and proportion of matched reads should be determined. Thus, the same inner function can be used for both to identify motif matches within a read. Then the number of motifs per read can be tabulated, where partial occurrences would be counted with a weight which is the fraction of the motif that was matched to the read.

HiyanZhang:
### Description

Given a motif (adaptor/contaminant) and a fasta file, compare the motif against all reads in the file and count the number of partial overlaps.

Fixes # (issue 19)

### Type of change

      to not work as expected)
- [ ] Documentation update

### Checklist

_or do not apply_.

      warnings
- [ ] I have commented my code in hard-to-understand areas
      works
- [ ] New and existing unit tests pass locally with my changes and I have not
      by the proposed changes

together.


htsinfer/count_occurrences_motifs.py
```diff
@@ -0,0 +1,78 @@
...
+             
+        overlaps_dict: collects all partially aligned substrings for a given number of partially aligned characters.
+    """
+    len_motif = len(motif)
+
+
+    if len_motif > len(read):
+        print('read: ', read)
+        print('motif: ', motif)
+        raise ValueError('the motif length needs to be shorter than the read length!')
```
mzavolan, line 28: It would be better to put all the information about the error in the error message. Can be done by creating an error class, in which the read and motif are variables.

htsinfer/count_occurrences_motifs.py
```diff
@@ -0,0 +1,78 @@
...
+
+
+def count_overlapping_motif_in_read(motif, read, overlap_cuttoff_ratio ):
+    """ Given a motif and a read, compare the motif against all substrings of the read to search for (partial) overlaps.
+    
+    Args:
+        motif: a short string, e.g., 'TTCTCCTAAGAATTGATTGAT'
+        read: a long string, whose length needs to be greater than that of the motif, e.g., 
+                    'TTCTCCTTTCTCCTAAGAATTGATTGATTCTCCTAAGAATTGATTGATAAGAATTGATTGAT'
+        overlap_cuttoff_ratio: a float between 0 and 1. Only overlaps larger than len(motif) * overlap_cuttoff_ratio will be taken into account.
```
mzavolan, line 15: This is too general. An 'occurrence' of a motif is a perfect match.

uniqueg:
Not quite what the specs describe, so better to reimplement from scratch.

PR #52, opened 2020-11-09
Issue_21_Identify_potential_barcode_k-mers_Ji_Hoon

### LINKED ISSUE #21: Identify potential barcode k-mers
Barcodes should consist of a small set of k-mers with the same length and a very similar position-dependent frequency of occurrence within reads. From the motif enrichment function we will get a list of strings that are considered significantly enriched among the reads. Another function will also implement computing the position-dependent frequency of occurrence of a k-mer in a set of input reads. We will use these two pieces of information, as well as the reads themselves to identify a set of motifs of the same length that together are represented across a minimal fraction of the reads and therefore could represent library barcodes. Inputs:  set of motifs (of potentially different lengths), set of reads. The function will first construct position-dependent frequency distributions. These will be compared pairwise across motifs of the same length. The list of pairwise distances will be traversed from smallest to largest, until a maximum threshold is reached, building sets of k-mers of the same length and with similar position-dependent frequencies. The coverage of reads by each set will be determined. The k for which the coverage satisfies a min threshold is selected. If multiple k's satisfy the condition, the largest one is chosen.

buffalo0124:
### Description

Barcodes should consist of a small set of k-mers with the same length and a very similar position-dependent frequency of occurrence within reads. Given a set of k-mers, compute the similarity of their position-dependent frequencies of occurrence. Determine the value of k and the precise set of k-mers that can explain most sequences in the library.

Fixes #21 

### Type of change

      to not work as expected)
- [ ] Documentation update

### Checklist

_or do not apply_.

      warnings
      works
      by the proposed changes

together.


uniqueg:
Closed as it does not implemented the desired functionality. See more detailed feedback in separate email (will be sent soon).

PR #53, opened 2020-11-10
Issue 18: Construct position-dependent frequency profiles

### LINKED ISSUE #18: Construct position-dependent frequency profiles
Given a set of motifs and a set of sequences, construct a position-dependent frequency profile of occurrence of the motifs within reads. This can be done by scanning all the reads with a particular motif, looking for perfectly compatible matches. That is, if the motif occurs internally in the read, it should be found precisely, and if the motif occurs at the 3' or 5' end of the reads it should be matched perfectly over the bases that are present. Record all the positions in the reads where the motif occurs and construct a histogram. If the motif occurs more than once in a read, count it fractionally towards each of those positions (e.g. if the motif occurs at 2 positions, each of these positions gets a count of 1/2). The simple implementation of this procedure will be slow (nr motifs * nr reads * read length), so make use of multiprocessing to speed things up.

mariapau189:
### Description

I wasn't able to solve the issue. Some of the small steps work but the output is not as expected. I solved the warnings but still only get the names of the sequences and not the positions of the motifs. Though the time fo the hackaton is over, if you have some time to comment on this I would appreciate it. Maybe I can try to make it work as I do my project. 

Fixes # 18

### Type of change

- [x ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
      to not work as expected)
- [ ] Documentation update

### Checklist

_or do not apply_.

- [x ] I have performed a self-review of my own code
- [x ] My code follows the existing coding style, lints and generates no new
      warnings
- [ ] I have commented my code in hard-to-understand areas
      works
- [ x] New and existing unit tests pass locally with my changes and I have not
      by the proposed changes

together.


uniqueg:
Closed because of some errors, wrong formatting/packaging. Probably not very useful.

PR #59, opened 2021-02-19
feat: organism and adapter count

rohank63:
#### infer_organism : 

> Added confidence logic.
> For every run creating a file for user with count match percentage for all organisms.
> Added more input parameters : 
>    - min_match :  minimum match percentage that top organism needs to have.
>    - factor : factor by which first organism is greater than the second. 

   
### Type of change

      to not work as expected)
- [X] Documentation update

### Checklist

- [ ] My code follows the existing coding style, lints and generates no new
      warnings
      works
      by the proposed changes


uniqueg:
Thanks a lot @rohank63 - this is a big step forward :)

There is still some work to do though. Apart from the specific comments that I will add in my review in due time, there are also a couple of more general issues to be addressed first:

* The commit history, when kept in good order, allows us to track the development of an application. When applying things like Conventional Commits, this further allows us to automate things like versioning, publishing releases, creating/updating changelogs etc. which not only saves us a lot of time as developers, but also makes it easier to users to see what changes between versions. It is therefore not only important to have well-written but also _semantic_ commits. The way we typically do that (and as outlined [here](https://github.com/elixir-cloud-aai/elixir-cloud-aai/blob/dev/resources/contributing_guidelines.md)) is by having people work on individual issues in dedicated feature branches. We can then leave them to commit however they want, quickly fix things, respond to PRs etc. without worrying too much about the count, structure, language of commits. Once a PR is ready, commits are then squashed into a single structured/semantic commit that gets added to the default branch (e.g.: `feat: add organism inference`). But this strategy does not work if you mix _different_ issues into a single PR like you did here. One option to fix this is by doing an interactive rebase to restructure a feature branch (and PR) into one or more meaningful commits (two in this case, one for the organism and one for the 3' adapter feature) that can then be added on top of the default branch in a rebase merge. However, I'm afraid the commit range of this branch might already be too convoluted to do that (it will for sure be if any one of your commits tackled both of these functionalities at the same time!). The other option is to create new branches, one for each feature, and roll out your commits to each of them manually, breaking up commits that tackled both features. Both of these branches should be created off the current default branch, so that no major merging conflicts will arise (there will be some when trying to merge the second functionality, but those should be easy to fix). It probably makes sense to start with a branch for the organism inference, as you were probably at least almost done with it by the time you started work on the adapter determination.
* In order for PRs to be ready for review, CI tests should all pass. This goes particularly for linting issues of which there are still many around. I would strongly recommend you to use an editor that allows you set up linting plugins that will inform you about linting issues _while you write the code_. Not only will this avoid such problems in the future, but gradually you will learn how to write code that pretty much always lints just fine (and thus improve your coding style).
* In order to ensure that _all_ code works as expected and (even more important) avoid code regression in the future, every line of code added to the code base should be unit tested. Exceptions to that rule should be kept to an absolute minimum and only for very good reasons if we decide that code cannot be meaningfully tested.

So, please, before I start adding more extensive reviews on the actual code, please:

- [ ] Prepare one PR for each functionality (organisms & 3' adapters) by creating new feature branches off the current default branch (which includes any changes that were introduced since last summer/fall) and adding your work on top of them
- [ ] Make sure all your code lints with the configured linters (`flake8` and `pylint`)
- [ ] Provide meaningful and complete unit tests for all desired functionality and that cover every line of added code; if you really find that some code is untestable, please describe why you think so and add references to the affected code
- [ ] Make sure Travis CI branch checks pass (don't worry about the PR test, that's an issue with the Travis CI config that I will fix later); they should if linting and testing are addressed

uniqueg:
Oh, one more specific issue I can already tell you from looking at the Travis CI logs: it is full of log messages saying something like "Processing read". Given that we are dealing with files that may have thousands or millions of reads, that is really not very reasonable. As it doesn't even give any meaningul information (which read? ID? Number?) this is not even a useful debug message. It's certainly good practice to print log messages, but be careful when putting them in loops, otherwise logs can very quickly become excessively big. Also make sure to log with as much information as possible and use the most appropriate log levels (`DEBUG`, `INFO`, `WARNING`, `ERROR` and `CRITICAL`) for each log message so that consumers can play with the verbosity levels depending on what they are looking for.

PR #60, opened 2021-02-22
feat : adapter count

rohank63:
### Description

      to not work as expected)
- [X] Documentation update

### Checklist

_or do not apply_.

      warnings
      works
      by the proposed changes


rohank63:
Hi @uniqueg ,
As you mentioned, I have created another branch by the name adapter for the feat: infer_adapter.
Added more unit tests. Coverage is now 100 %
Fixed linting bugs.
However 2 linting type errors are still there in the build. I don’t know what they stand for
Rest is taken care of.

.travis.yml
```diff
@@ -34,3 +34,4 @@ jobs:
         # integration tests
         - htsinfer --help
         - htsinfer --debug -f1 tests/test_files/first_mate.fastq
+        - htsinfer --debug -f1 tests/test_files/SRR13496438.fastq.gz
```
uniqueg, line 37: maybe it would be good to rename the file to give it a more semantic meaning. what is this testing? something like `human_real_world_sample_10000_records.fastq.gz`

htsinfer/htsinfer.py
```diff
@@ -48,6 +49,20 @@ def parse_args(
             "record; set to 0 to process entire file(s)"
         )
     )
+    parser.add_argument(
```
uniqueg, line 52: Seems to me that these options are still from the organism functionality. It's important that parameter names are non-ambiguous, and we will have similar names for adapters and organisms, so better name them something like `adapter-designation-min-match-percentage`, `adapter-designation-frequency-ratio`, `organism-designation-min-match-percentage`, `organism-designation-frequency-ratio`. But of course keep the `adapter-` parameters for this PR and the `organism-` ones for the organism PR. Also, note that by convention, CLI parameters should use dashes instead of underscores. The `argparse` package will automatically replace these with underscores for use within Python. Also, for CLI parameter names preceded by a _single_ dash, by convention, there should be only a _single_ character. But these shorthand notations should only be supplied for very common parameters, like `--help | -h` and similar. I would _not_ provide them here.

htsinfer/htsinfer.py
```diff
@@ -48,6 +49,20 @@ def parse_args(
...
+        '-mm', '--min_match',
+        metavar="FLOAT",
+        type=float,
+        default=10,
+        help="minimum match percentage that organism needs to have"
```
uniqueg: This is not a very clear/precise description (and neither is the one for factors). To make sure people really understand how to use these parameter, something like this would be better (for adapters, slightly different for organisms of course):

> Minimum percentage of reads that contain a given adapter in order for that adapter to be considered as the resulting adapter. If no adapter is found more frequently than the specified number (in percent), `null` is returned in the JSON result to indicate that no adapter could be confidently identified.

And:

> The minimum frequency ratio between the first and second most frequent adapters in order for an adapter sequence to be returned in the JSON result. If the frequency ratio is less than the specified value, `null` is returned in the JSON result to indicate that no single adapter could be confidently identified.

uniqueg: It's okay to have longish descriptions. It's much preferable over users not being able to understand how they can use a tool. In that regard: Make sure that help messages are consistent with regard to their structure/style/format. As far as I'm concerned, it's totally okay (probably better) to have help messages that read like proper sentences, start with capital letters and contain punctuation, including a closing `.`. It's up to you though, as long as the help messages of _all_ parameters are consistent in their style.

uniqueg: Another note: the final output should be a JSON-formatted string, not just a printed Python dictionary (as parsing Python dictionaries is not as straightforward as JSON and JSON is used much more widely and language-independent). That's why I included `null` in the description and not `None`. Of course in your code you would return objects as dictionaries (or possibly instances of model classes that _look_ like dictionaries but allow validation of return objects). That stuff can be done later, in a different issue/PR, but perhaps good that (a) you are already aware of that and (b) that the descriptions already match what we want to have.

htsinfer/htsinfer.py
```diff
@@ -148,6 +163,12 @@ def main() -> None:
         file_2=args.file_2,
         organism=args.organism,
     )
+    results['adapters'] = infer_adapter.infer(
+        file_1=args.file_1,
+        file_2=args.file_2,
+        min_match=args.min_match,
+        factor=args.factor
+    )
```
uniqueg, line 193: Also not something to implement now, but in a future issue/PR: the results object (currently a simple dictionary) should probably be a model class in order to allow proper validation. This can be done very easily with the Pydantic package, which makes use of Python's type hinting system to define model classes with very little overhead. This makes the code much safer and will also make it easier to document/understand the code as it is getting more complex. I think Pydantic model class instances also have methods that allow direct export to JSON, so that's useful to tackle the issue with JSON results mentioned above.

Note that when defining model classes, these should be hierarchical, so you'd have a model `Results()` that contains, eg., a property `adapter` of type `ResultsAdapter()` - which is in itself a model class. The FOCA package has some examples ([API docs](https://foca.readthedocs.io/en/latest/modules/foca.models.html), [actual code](https://github.com/elixir-cloud-aai/foca/blob/dev/foca/models/config.py)).

Please write an issue out of this and the previous command on representing result objects as Pydantic model classes. I think the JSON export can be handled together with that issue as Pydantic models can indeed be exported to JSON directly (https://pydantic-docs.helpmanual.io/usage/exporting_models/#modeljson).

htsinfer/infer_adapter.py
```diff
@@ -0,0 +1,203 @@
...
+    Returns:
+        bool : whether it satisfies confidence score or not.
+    """
+    if adapters_df.iloc[0]['Count %'] < min_match:
+        return False
+    if adapters_df.iloc[1]['Count %'] != 0:
+        ratio = adapters_df.iloc[0]['Count %']/adapters_df.iloc[1]['Count %']
+        if ratio < factor:
+            return False
+    return True
```
uniqueg: Please add the next file `tests/.DS_Store` to the `.gitignore` file as it should not be version-controlled, I think.

tests/test_unit_infer_adapter.py
```diff
@@ -0,0 +1,86 @@
+"""Unit tests for infer_adapter.py"""
```
uniqueg: Tests look good as far as I can tell. Will have another look once all other issues (and particularly those in the main adapter inference module) are resolved.

htsinfer/infer_adapter.py
```diff
@@ -0,0 +1,203 @@
+"""Infer adapter sequences present in reads."""
+
```
uniqueg, line 2: Please remove empty line, should be only one

htsinfer/infer_adapter.py
```diff
@@ -0,0 +1,203 @@
...
+
+from functools import partial
```
uniqueg, line 4: Please follow [PEP 8](https://www.python.org/dev/peps/pep-0008/#imports) for structuring imports and please, within each group of packages, sort packages alphabetically.

For example:
```Py
from functools import partial
import gzip
import logging
from typing import (Dict, List, Tuple)

import ahocorasick as ahc
from Bio.SeqIO.QualityIO import FastqGeneralIterator
import pandas as pd
```

I would also recommend not to import entire modules, but only import what you need. I know it's pretty standard to import `pandas` like this, but since you only need a _single_ object from the package, it's really much better to import only that. Same is true for `ahocorasick`

htsinfer/infer_adapter.py
```diff
@@ -0,0 +1,203 @@
...
+import gzip
+from typing import (Dict, List, Tuple)
+import pandas as pd
+import ahocorasick as ahc
+
+from Bio.SeqIO.QualityIO import FastqGeneralIterator  # type: ignore
+
+LOGGER = logging.getLogger(__name__)
+
+
```
uniqueg: Please remove empty line, should be only one

htsinfer/infer_adapter.py
```diff
@@ -0,0 +1,203 @@
...
+ADAPTERS = [
```
uniqueg: These should be read from a file that is supplied with the package, similar to the transcripts file

uniqueg: Also, consider using an enum for this, specifically an `IntEnum`

htsinfer/infer_adapter.py
```diff
@@ -0,0 +1,203 @@
...
+    ('TCGTCGGCAGCGTCA', 24),
+    ('TGGAATTCTCGGGTG', 25),
+    ('UCGUAUGCCGUCUUC', 26),
+    ('AGATCGGAAGAGCGT', 27)
+]
+
+
+def make_aho_auto(
+    adapters: List[Tuple[str, int]]
+):
```
uniqueg: Specify return type

htsinfer/infer_adapter.py
```diff
@@ -0,0 +1,203 @@
...
+    """Adding all adapters names into trie"""
```
uniqueg: Please write full Google-style docstrings (https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html) and make sure to use imperative, e.g., "Return trie of adapters."

htsinfer/infer_adapter.py
```diff
@@ -0,0 +1,203 @@
...
+    LOGGER.debug("Creating trie")
```
uniqueg: Trie of what?

htsinfer/infer_adapter.py
```diff
@@ -0,0 +1,203 @@
...
+    file_1: str,
+    file_2: str = None,
+    max_records: int = 100000,
+    min_match: float = 10,
+    factor: float = 2
+) -> Tuple[str, str]:
+    """Infers adapter information for one or two fastq files.
+
+    Args:
+        file_1 (str) : File path to read/first mate library.
```
uniqueg: Please do _not_ include type annotations in docstrings, they are already in the function headers. Having them in two places is likely to lead to divergence in the future. The code base is always the actual source of truth and preferable to comments whenever possible. That's why it's important to write clear code with self-explanatory names etc

htsinfer/infer_adapter.py
```diff
@@ -0,0 +1,203 @@
...
+        file_2 (str) : File path to second mate library.
+        max_records: Limit processing to the indicated number of records,
+            starting from the first record.
+        min_match (float) : minimum match percentage that first adapter needs
+        to have.
+        factor (float) : factor by which first adapter is greater than the
+        second adapter.
+
+    Returns:
+        Type of Adapter that is present in library.
```
uniqueg: It returns the adapter _sequence_, not its _type_ (whatever that may be). Also, "adapter" rather than "Adapter"

htsinfer/infer_adapter.py
```diff
@@ -0,0 +1,203 @@
...
+    """Infers adapter information for one or two fastq files.
+
+    Args:
+        file_1 (str) : File path to read/first mate library.
+        file_2 (str) : File path to second mate library.
+        max_records: Limit processing to the indicated number of records,
+            starting from the first record.
+        min_match (float) : minimum match percentage that first adapter needs
+        to have.
+        factor (float) : factor by which first adapter is greater than the
```
uniqueg: Please be consistent with docstrings. Capitalize first letter and add punctuation in _all_ cases. Also, of course you should look at the comment above for the corresponding CLI parameters. You should reuse the extensive and precise descriptions.

htsinfer/infer_adapter.py
```diff
@@ -0,0 +1,203 @@
...
+
+    Returns:
+        Type of Adapter that is present in library.
+    """
+    # Process file 1
+    LOGGER.debug(f"Processing file 1: {file_1}")
+    result_1 = process_fastq_file(file_1, max_records, min_match, factor)
+
+    # Process file 2
+    result_2 = "not_available"
```
uniqueg: Please do not use a string as an unavailable result, but rather something like `None`

htsinfer/infer_adapter.py
```diff
@@ -0,0 +1,203 @@
...
+    ('TCGGACTGTAGAACT', 22),
+    ('TCGTATGCCGTCTTC', 23),
+    ('TCGTCGGCAGCGTCA', 24),
+    ('TGGAATTCTCGGGTG', 25),
+    ('UCGUAUGCCGUCUUC', 26),
+    ('AGATCGGAAGAGCGT', 27)
+]
+
+
+def make_aho_auto(
```
uniqueg: I think it would make the module more readable if this function was added below the FASTQ processing function

htsinfer/infer_adapter.py
```diff
@@ -0,0 +1,203 @@
...
+    LOGGER.debug("Creating trie")
+    trie = ahc.Automaton()
+    for (adapter, tag) in adapters:
+        trie.add_word(adapter, (tag, adapter))
+
+    trie.make_automaton()
+    return trie
+
+
+TRIE = make_aho_auto(ADAPTERS)
```
uniqueg: Constants should be defined at the top

htsinfer/infer_adapter.py
uniqueg: Better to use else here, as you'd have to assing twice to `result_2` if `file_2` is available

htsinfer/infer_adapter.py
```diff
@@ -0,0 +1,203 @@
...
+    Returns:
+        Type of Adapter that is present in library.
+    """
+    # Process file 1
+    LOGGER.debug(f"Processing file 1: {file_1}")
+    result_1 = process_fastq_file(file_1, max_records, min_match, factor)
+
+    # Process file 2
+    result_2 = "not_available"
+    if file_2:
```
uniqueg: If the default of `file_2` is `None` and not an empty string, this is not the proper way to test. See the second bullet point here: https://www.python.org/dev/peps/pep-0008/#programming-recommendations

Might be a good idea for you to read the entire PEP 8, it's really really important

htsinfer/infer_adapter.py
```diff
@@ -0,0 +1,203 @@
...
+        LOGGER.debug(f"Processing file 2: {file_2}")
```
uniqueg: Better (more descriptive) to say "mate 2 file"

uniqueg: Btw, please write an issue to remind us to check whether searching for adapters in second mates is the same as in first mates, or whether they are possible reversed/complemented. We can probably only check that once we have implemented the read orientation inference and have all the proper test cases.

htsinfer/infer_adapter.py
```diff
@@ -0,0 +1,203 @@
...
+        min_match (float) : minimum match percentage that first adapter needs
+        to have.
+        factor (float) : factor by which first adapter is greater than the
+        second adapter.
+
+    Returns:
+        Type of Adapter that is present in library.
+    """
+    # Process file 1
+    LOGGER.debug(f"Processing file 1: {file_1}")
```
uniqueg: Better to use `mate 1 file` or `first mate file`

htsinfer/infer_adapter.py
```diff
@@ -0,0 +1,203 @@
...
+    """
+    _open = partial(
+        gzip.open, mode='rt'
+    ) if file.endswith(".gz") else open
+
+    try:
+        LOGGER.debug("Opening file...")
+        with _open(file) as _file:  # type: ignore
+
+            adapter_counts: Dict[str, float] = {}
```
uniqueg: Better to initialize a `defaultdict` and set all counts to zero: https://www.geeksforgeeks.org/defaultdict-in-python/

htsinfer/infer_adapter.py
uniqueg: Guess you can load the adapters file in this function

htsinfer/infer_adapter.py
```diff
@@ -0,0 +1,203 @@
...
+            records: int = 0
+            total_count: int = 0
+            LOGGER.debug("Processing reads")
+            for record in FastqGeneralIterator(source=_file):
+                # Get next read
+                read = record[1]
+
+                # Searching for adapters in read
+                for _, (_, key) in TRIE.iter(read):
+                    if key in adapter_counts:
```
uniqueg: Refactor to make use of `defaultdict`

htsinfer/infer_adapter.py
```diff
@@ -0,0 +1,203 @@
...
+    try:
+        LOGGER.debug("Opening file...")
+        with _open(file) as _file:  # type: ignore
+
+            adapter_counts: Dict[str, float] = {}
+            records: int = 0
+            total_count: int = 0
+            LOGGER.debug("Processing reads")
+            for record in FastqGeneralIterator(source=_file):
+                # Get next read
```
uniqueg: Trivial comment, can omit

htsinfer/infer_adapter.py
```diff
@@ -0,0 +1,203 @@
...
+
+                # Searching for adapters in read
+                for _, (_, key) in TRIE.iter(read):
+                    if key in adapter_counts:
+                        adapter_counts[key] += 1
+                    else:
+                        adapter_counts[key] = 1
+                    total_count += 1
+
+                # Update records
```
uniqueg: Trivial comment, can omit

htsinfer/infer_adapter.py
```diff
@@ -0,0 +1,203 @@
...
+                records += 1
```
uniqueg, line 117: Better to move up with `read = record[1]`, no?

htsinfer/infer_adapter.py
```diff
@@ -0,0 +1,203 @@
...
+
+                # End if requested records processed
+                if max_records and records >= max_records:
+                    break
+
+            if total_count == 0:
+                return "NA"
```
uniqueg: Again, no strings to represent not available

uniqueg: Should use `None` instead which will be translated to `null` when exported to JSON

htsinfer/infer_adapter.py
```diff
@@ -0,0 +1,203 @@
...
+                    total_count += 1
+
+                # Update records
+                records += 1
+
+                # End if requested records processed
+                if max_records and records >= max_records:
+                    break
+
+            if total_count == 0:
```
uniqueg: All of this and the following code does not need the open input file object, so `with` statement should stop here and OSError catch statement should follow first

htsinfer/infer_adapter.py
uniqueg: Why not define all of these before openeing the file object? Just put the code in the `with` context manager that you really need the file open for

htsinfer/infer_adapter.py
```diff
@@ -0,0 +1,203 @@
...
+                return "NA"
+
+            # Calculating Percentage
+            for i in adapter_counts:
```
uniqueg: Actually, this should be divided by records, not total counts. Otherwise we might pick adapters that occur only once or so. In fact, if only a single adapter matches once and all the others don't it would be returned. Not very robust

uniqueg: Also, you can probably represent this as a list comprehension

htsinfer/infer_adapter.py
```diff
@@ -0,0 +1,203 @@
...
+                records += 1
+
+                # End if requested records processed
+                if max_records and records >= max_records:
+                    break
+
+            if total_count == 0:
+                return "NA"
+
+            # Calculating Percentage
```
uniqueg: I'm thinking, since you'll need this (and the following) code also in the organism inference, I would put this in a separate utilities module and import. Calculating the percentage would be one function, converting dict to data frame a second, writing a Pandas data frame to a file a third, defining the adapter/organism to return a fourth

uniqueg: Also, not sure if all of these functions should really be called in `process_file()` function, since what you process in the file is really only what you do when the file object is open. Might make more sense to put them in `infer()`, at least the output file writing and decision/designation part.

htsinfer/infer_adapter.py
```diff
@@ -0,0 +1,203 @@
...
+            # Converting dictionary into dataframe
+            adapters_df = pd.DataFrame(adapter_counts.items())
+            adapters_df.columns = ['Adapter', 'Count %']
+            adapters_df = adapters_df.sort_values(
+                by='Count %', ascending=False
+                ).reset_index(drop=True)
+            LOGGER.debug(f"Creating {file}_adapters_count.csv")
+            adapters_df.to_csv(f"{file}_adapters_count.csv")
+
+            # Checking confidence score
```
uniqueg: I think confidence is not a good term to use in the actual code as it kinda implies its use in math, and we are certainly not calculating any confidence. In fact, the same is true for inference, we are not inferring anything, we are just counting :)

htsinfer/infer_adapter.py
```diff
@@ -0,0 +1,203 @@
...
+            if confidence(adapters_df, min_match, factor):
+                result = adapters_df.iloc[0]['Adapter']
+            else:
+                result = "NA"
```
uniqueg: As above

htsinfer/infer_adapter.py
```diff
@@ -0,0 +1,203 @@
...
+            return result
+
+    except OSError:
+        LOGGER.error(f"Invalid input file '{file}'")
+        return "invalid_file"
+
+
+def confidence(
```
uniqueg: As mentioned above: put in separate model as it can be reused in organism module. But think of a more precise name

uniqueg (changes requested):
Overall looks good, but still quite a lot of stuff to make it really elegant/pretty :)

PR #78, opened 2022-05-18
fix(orientation): handle sources

uniqueg:
### Description

- library source in read orientation inferral is taken from previous inference of library source


uniqueg:
Closing for #79

PR #90, opened 2022-07-26
test: add unit tests for library source inference

rohank63:
### Description

Please include a summarized description of the change(s) and which issue(s)
this pull request addresses, if any. Please include relevant motivation and
context, unless already duly described in an associated issue or resource.
Make sure to list any dependencies that are required for this change.

Fixes # (issue)

### Type of change

- [X] Extensive Unit Testing
- [X] Documentation update

### Checklist

_or do not apply_.

      warnings
      works
      by the proposed changes


htsinfer/get_library_source.py
```diff
@@ -270,26 +271,32 @@ def get_source_expression(
...
+            if total_tpm != 0:
+                dat_agg.tpm = dat_agg.tpm / total_tpm * 100
+
+            # return as dictionary
+            return dat_agg.sort_values(["tpm"], ascending=False)
+
+        except (AttributeError, ValueError, OSError) as exc:
+            raise FileProblem(
+                "Could not process file: {_file}"
+            ) from exc
```
uniqueg: Not a fan of wrapping a ton of code in some `try` - `except` block and then catching a bunch of errors. The resulting information for the user will invariably vague, just as vague as if errors weren't caught at all.

Moreover, the data is already loaded at this point, so whatever happens, it's certainly not really a `FileProblem` (in the sense that something went wrong with opening the file), but rather some sort of `ValueError` - yes, the file is there, it could be opened, but it doesn't quite look as expected.

I'd suggest the following: If you have any specific ideas of what might go wrong with a specific line or small block of code (e.g., the data frame may be empty), then catch just these problems, case by case. If not, no problem then. Maybe in the future someone will report an error/problem with the code, and then we can still catch errors accordingly.

In any case, please remove that big, catch all `try` - `except` block again.

If not,

tests/files/abundance.tsv
```diff
@@ -0,0 +1 @@
+organism	taxon_id
```
uniqueg, line 1: That's weird content for a file called `abundance.tsv`. One would expect a (small) transcript abundance table like the ones that `kallisto` produces.

tests/utils.py
```diff
@@ -43,9 +43,10 @@
 FILE_ORIENTATION_U = TEST_FILES_DIR / "orientation_u.fastq"
 FILE_2000_RECORDS = TEST_FILES_DIR / "2000_records.fastq"
 FILE_SINGLE = TEST_FILES_DIR / "single.fastq"
+FILE_SOURCE_FRUIT_FLY = TEST_FILES_DIR / "fruit_fly.fastq"
 FILE_SRA_SAMPLE_1 = TEST_FILES_DIR / "sra_sample_1.fastq"
 FILE_SRA_SAMPLE_2 = TEST_FILES_DIR / "sra_sample_2.fastq"
-FILE_TRANSCRIPTS = TEST_FILES_DIR / "transcripts.fa"
+FILE_TRANSCRIPTS = TEST_FILES_DIR / "transcripts.fasta"
```
uniqueg, line 50: I'm confused. It seems you added a new file `transcripts.fasta`. But there was (and still _is_) already a file `transcripts.fa`. How do they differ?

If they are both valid but functionally different (i.e. they are used to test different behavior), then rename them so that it's clear what they are. If they are the same, just keep one. If the old one is invalid, also just keep one.

tests/test_get_library_source.py
```diff
@@ -68,3 +77,201 @@ def test_evaluate_paired(self, monkeypatch, tmpdir):
...
+            tmp_dir=tmpdir,
+            out_dir=tmpdir,
+        )
+        results = test_instance.evaluate()
+        assert results == ResultsSource(
+            file_1=Source(),
+            file_2=Source()
+        )
+
+    def test_evualte_file_problem_transcripts(self, tmpdir):
```
uniqueg: Typo in `evaluate`

uniqueg (changes requested):
Looking pretty good, thanks a lot! Just a few minor comments

tests/utils.py
rohank63: The reason why I added the new one, as I was checking the read_source for different organisms the older file was not giving somehow proper tests. So then I added my original transcripts file and it worked, and the fact that I didn't remove the earlier one, just to bring into your notice.

Next commit, I will remove that.

tests/files/abundance.tsv
rohank63: It was intentional, to make it empty and not include any data/table because when we run the kallisto_quantification , if it fails or (in somecases in the past, not able to recall which usecase ) , it will still produce an abundance.tsv which will be empty , so to cover that aspect I made it like that. 

Still, let me know your comments on this.

htsinfer/get_library_source.py
rohank63: Yeah, you are right, I agree to that . File is loaded so Raising File Problem may not be best Raise. 

But Why I added  a try - catch was to handle a testcase ( I have described below : in abundance.tsv section ). 
If the file is empty, without the try-catch code, it will give error. 

So here, I can do two things : 
1. Reduce the Try-Catch Block and Raise something else instead of FIleProbelm.
2. Add a check, to handle if data frame is empty . 

I think the first one makes more sense to me, as it can handle more corner cases which I might be missing , but can arise in the future.

Still. lemme know your thoughts on this.

uniqueg: I agree that it's better to `try` - `catch` than add a conditional to check explicitly if the file is empty. But you should then identify the (first) statement that raises an error on an empty data frame and then only wrap that one in `try` - `catch`. But let's not just catch any kind of error that might occur but that we can't think of now. If we find more of these, raised by the same or other statements, we can still catch those errors and react accordingly. But at any given point, let's only catch those errors that we _know_ can/will happen

tests/files/abundance.tsv
uniqueg: Then call it something like `FILE_EMPTY_TABLE` (for the variable in the tests) and `empty_table.tsv` (for the filename itself). Then we won't have this sort of confusion.

htsinfer/get_library_source.py
```diff
@@ -270,6 +271,11 @@ def get_source_expression(
                 "Could not process file: {_file}"
             ) from exc
 
+        if dat.empty:
+            raise ValueError(
```
uniqueg: What is the possible cause and consequence of this? No chance of recovering from this (e.g., by setting the library source to `UNKNOWN` or whatever the appropriate state is)? Not quite sure whether raising an error is the best choice here.

If it _does_ turn out that raising an error is the best approach, then please list the exception and its cause in the docstring under `FileProblem`.

pylint.cfg
```diff
@@ -1,4 +1,4 @@
 [MESSAGES CONTROL]
-disable=C0330,I1101,R0801,R0902,R0903,R0913,R0914,W1202,W1203,W1510
+disable=C0330,I1101,R0801,R0902,R0903,R0913,R0914,W1202,W1203,W1510,E1101,C2801
```
uniqueg: See other PR.

uniqueg (changes requested):
Very minor comments, thanks a lot :)

uniqueg (approved):
Now let's see how we merge this with the other open PR from @BorisYourich :)

PR #91, opened 2022-08-25
test: add unit tests for library orientation 

### LINKED ISSUE #83: Implement unit tests for read orientation inference

Currently, comprehensive unit tests are missing for module [`htsinfer.get_read_orientation`](https://github.com/zavolanlab/htsinfer/blob/dev/htsinfer/get_read_orientation.py).

Implement unit tests for the indicated module. The goal is not only to achieve 100% test coverage, i.e., run each line of code in a given test run, but to test the intended/expected behaviors of _each individual method/function_ separately and comprehensively. Available unit tests for other modules (e.g., `htsinfer.py`, `get_read_layout`) can serve as an example.

N/A

**Additional context**

Note that in order to pass CI checks, coverage checks for the module have been excluded in [`setup.cfg`](https://github.com/zavolanlab/htsinfer/blob/dev/setup.cfg). Please remove the module from the corresponding config section before filing the pull request.

rohank63:
Resolves #83 

### Type of change

- [X] Extensive Unit Testing
- [X] Documentation update

### Checklist

_or do not apply_.

      warnings
      works
      by the proposed changes


htsinfer/get_read_orientation.py
```diff
@@ -446,7 +446,7 @@ def process_single(
                                 StatesOrientation.stranded_forward
                             )
 
-        except OSError as exc:
+        except (OSError, ValueError) as exc:
```
uniqueg: Can you please explain under what conditions a `ValueError` is raised? I mean I'm sure that it can happen with all that code in the loop, especially if the file is not a valid SAM file. But it might make more sense to catch the errors separately and give a more informative error message, becaue `Failed to open SAM file` is probably not quite appropriate for `ValueError`. Same comment applies to the change below.

Also, it seems that I forgot to document the raised exceptions in the docstring. After we have discussed what to do here, please add a section `Raises:` to the docstring and list the exceptions that may be raised by this method and the circumstances under which they are raised.

pylint.cfg
```diff
@@ -1,4 +1,4 @@
 [MESSAGES CONTROL]
-disable=C0330,I1101,R0801,R0902,R0903,R0913,R0914,W1202,W1203,W1510
+disable=C0330,I1101,R0801,R0902,R0903,R0913,R0914,W1202,W1203,W1510,E1101,C2801
```
uniqueg: I think globally disabling linter messages should be the exception (as it is, there were already too many). And `C2801` and particularly `E1101` are not messages that we would want to globally disable. Can you please remove these here and add `# pylint disable=E1101` comments to the lines where you _really_ need them? Better of course to avoid them entirely.

tests/test_get_read_orientation.py
```diff
@@ -1,5 +1,8 @@
 """Unit tests for module ``get_read_orientation.py``."""
```
uniqueg, line 1: Commenting here because I can't comment on an empty file and I don't want to comment on a 170'000 lines file :)

* It doesn't really make sense to add a description and file extension to an empty file. Just call it `empty` and the variable `FILE_EMPTY`, it's shorter, more precise and makes more sense to reuse in other contexts.
* Why add `transcripts.fasta` (a big file!) again, when a compressed version is already in `data/transcripts.fasta.gz`? Couldn't you just use that? If not, for whatever reason, I think it would make sense to just use a small (max 10'000 lines, better 1'000) subset of the file.

tests/utils.py
```diff
@@ -67,6 +70,7 @@
 
 # helper classes
 class RaiseFileProblem:
+    """Raise ``FileProblem`` ."""
```
uniqueg: I know you didn't add this code, but it's a good opportunity to mention this:

I have recently found a way to write a generic exception raiser class. You can just pass it the exception that you want to raise. Reduces some duplicate code. Check here: https://github.com/zavolanlab/zarp-cli/blob/4b93643b8650656b18b9301e91413469bff23e6f/tests/utils.py#L6-L17

Maybe you want to add that and replace the usage of `RaiseFileProblem`, `RaiseMetadataWarning`, `RaiseOSError` and `RaiseValueError`? :)

uniqueg (changes requested):
Tests look good - thanks a lot :) Few comments, mostly minor, I think

htsinfer/get_read_orientation.py
rohank63: Value Error is raised in such scenarios, when a sam file is generated, but alignments are not found.

tests/test_get_read_orientation.py
rohank63: I did't add a new transcripts.fasta, I just took (transcripts.fa) from dev and compared to the transcripts.fasta that I had and renamed.
But surely, we can make a subset of that.

uniqueg: Okay, probably better to have a subset in the test files. We already have the complete file in `data/`

htsinfer/get_read_orientation.py
uniqueg: So if the file is empty? Or if the records are corrupt / not valid SAM? Either way, would be good to give a more accurate/informate error message, like `Input file empty` or `Not a valid SAM file`.

rohank63: Sure, I ll change it.

tests/test_get_read_orientation.py
rohank63: @uniqueg : I tried to do a subset, but thing is for infer_source for extensive testing I used some source organisms eg Fruit-fly (dmelanogaster). So if I take subset, I may miss the transcripts for that organism. One way is to hard find those transcripts then add as a subset along with hsapiens and some random. 
PS: I tried to do that, but for some reason, my system is crashing when I open transcripts.fasta.

uniqueg: Well, my issue is that there is a big (10 Mb) uncompressed file in `tests/files`, and this file was added in _this_ PR. And I don't think it's (a) necessary that we have such a big test file and (b) a compressed version of the file is _already_ in the repo (in `data/`). Why not use the file in `/data`?

And you didn't just rename the file. 170+k lines were added, 3+k lines were removed:
image

As for subsetting, you don't need to open the file. Just use a tool like `grep` to subset.

uniqueg (approved):

uniqueg:
Good stuff - thanks a lot @rohank63!

PR #92, opened 2022-09-29
refactor: pass conf instances, not individual args

### LINKED ISSUE #84: Refactor argument passing to class instances

Currently, arguments are passed to class constructors individually. This has several disadvantages:

- Arguments need to be described in multiple locations (e.g., CLI interface, `HtsIinfer` class, classes for each inference functionality)
- Adding new arguments to classes requires code changes in multiple modules
- Constructor interfaces become unwieldy; `pylint` messages that warn against unwieldy interfaces were globally disabled (specifically: `R0902: too-many-instance-attributes` and `R0903: too-many-arguments`)

- [x] Define a (Pydantic) configuration model `Args` with all command-line arguments in [`htsinfer/models.py`](https://github.com/zavolanlab/htsinfer/blob/dev/htsinfer/models.py); add appropriate defaults
- [x] Define a (Pydantic) configuration model `Config` in [`htsinfer/models.py`](https://github.com/zavolanlab/htsinfer/blob/dev/htsinfer/models.py) with properties `args` of type `htsinfer.models.Args` and `results` of type `htsinfer.models.Results`
- [x] Instantiate an instance of the model in the `HTSinfer` class constructor - which will then take only two arguments, `args` and `results`, both optional, with defaults being specified in `models.py`
- [ ] Pass arguments in bulk to `Args()` in `HtsInfer` instantiation in [`htsinfer/cli.py`](https://github.com/zavolanlab/htsinfer/blob/dev/htsinfer/cli.py)
- [ ] Redefine constructors of classes in all other modules (e.g., `GetLibSource`, `GetReadOrientation`) such that they only take a single argument, `config` of type `htsinfer.models.Config`; collapse separate classes for handling two or only one input file into one, where applicable (newer functionalities use only a single class) as there is no solid reason to keep these separate, again causing repetition of descriptions and sometimes code
- [x] Refactor calls to instantiate the modified classes in [`htsinfer/htsinfer.py`](https://github.com/zavolanlab/htsinfer/blob/dev/htsinfer/htsinfer.py) so that only a single argument `config` is passed

N/A

**Additional context**

Following the suggested solution still requires arguments to be described in two places: the model definition in [`htsinfer/models.py`]
https://github.com/zavolanlab/htsinfer/blob/dev/htsinfer/models.py) and the CLI definition in [`htsinfer/cli.py`](https://github.com/zavolanlab/htsinfer/blob/dev/htsinfer/cli.py). An ideal solution would re-use the definitions from `model.py` in model for `argparse` help messages in `cli.py`. However, this is likely not trivial and should probably constitute a separate issue.

### LINKED ISSUE #93: pytest issue on macOS Big Sur
**Describe the bug**
Os-dependent issue when testing the installation of development/testing dependencies.
Error is: `htsinfer.exceptions.StarProblem: Failed to create STAR index`
If solved then: `htsinfer.exceptions.StarProblem: Failed to generate STAR alignments`

**To Reproduce**

1. Use computer with macOS Big Sur and Miniconda(4.12.0) installation.
2. Follow the Installation tutorial for `htsinfer`.

**Expected behavior**
To pass all tests.

**Desktop (please complete the following information):**

- OS: macOS Big Sur

**Additional context**
The problem solution is trivial although not up to me to decide 
if it should be a part of the codebase.

**Solution**
In file `htsinfer/htsinfer/get_read_orientation.py` (dev branch)
Add `index_dir.mkdir(parents=True, exist_ok=True)` in line 255.
Add `Path(out_dir).mkdir(parents=True, exist_ok=True)` in line 319.

BorisYourich:
### Description

Refactor argument passing to inference classes (#84)
Fix `pytest` issue on MacOS (#93)

* Pydantic configuration models `Args` and `Config` were defined and implemented
* The individual arguments of the HtsInfer class were removed and only a single non-optional argument `config` was added
* All of the attributes except `run_id` and `state` were removed as they are identical to `config.args` class attributes
* The previously used `self.results` attribute has been replaced by `self.config.results` because of redundancy
* Tests were modified to work with the refactored codebase
* Directories for STAR alignment and genome indexing are created in the script (#93)
* Minor changes in the code where deemed useful e.g. variable `orientation` removed from get_read_orientation.evaluate() due to redundancy.

Fixes #84
Fixes #93

### Type of change

- [x] Documentation update

### Checklist

_or do not apply_.

      warnings 
      works
      by the proposed changes

together.


htsinfer/cli.py
```diff
@@ -265,13 +268,7 @@ def main() -> None:
...
         args = parse_args()
 
-        # set up logging
-        setup_logging(verbosity=args.verbosity)
-        LOGGER.info("Started HTSinfer...")
-        LOGGER.debug(f"CLI arguments: {args}")
-
-        # determine library metadata
-        hts_infer = HtsInfer(
+        arguments = Args(
```
uniqueg: It would be nice to make use of `dest` in the CLI argument definitions to define the names for each variable rather than mapping the more descriptive names from the CLI args to the more concise ones used interally. That would allow us to pass arguments to `Args` through unpacking `args`.

For example:

```py
    parser.add_argument(
        "--output-directory",
        dest="out_dir",  # <-- this line is new
        default=Path.cwd() / 'results_htsinfer',
        type=lambda p: Path(p).absolute(),
        metavar="PATH",
        help="path to directory where output is written to",
    )
```

Now, if we do that for all of the args, we could do something like this (untested):

```py
args_dict: Dict = vars(args)
del args_dict["paths"]
arguments = Args(path_1=args.paths[0], path_2=args.paths[1], **args_dict)
```

Or even more concisely, we could do:

```py
config = Config(
   args=Args(path_1=args.paths[0], path_2=args.paths[1], **args_dict),
   results=Results(),
)
```

htsinfer/models.py
```diff
@@ -325,3 +327,120 @@ class Results(BaseModel):
...
+    Args:
+        args: Container class for CLI arguments.
+        results: Container class for aggregating results
+                from the different inference functionalities.
+    Attributes:
+        args: Container class for CLI arguments.
+        results: Container class for aggregating results
+                from the different inference functionalities.
+    """
+    args: Args
```
uniqueg: Might as well add defaults:

```py
args: Args = Args()
results: Results = Results()
```

pylint.cfg
```diff
@@ -1,4 +1,4 @@
 [MESSAGES CONTROL]
-disable=C0330,I1101,R0801,R0902,R0903,R0913,R0914,W1202,W1203,W1510
+disable=C0330,I1101,E1101,R0801,R0902,R0903,R0913,R0914,W1202,W1203,W1510
```
uniqueg: Hmmm, not quite sure where there's something dynamic happening in the changed code in terms of class member access. The Pydantic models are defined and everything seems to be typed. Does this show up in multiple places? If only in one or few, better to use `# pylint: disable=E1101` on the offending lines. I suspect that this can be avoided altogether.

htsinfer/htsinfer.py
```diff
@@ -37,126 +35,35 @@ class HtsInfer:
...
-            else self.tmp_dir / self.transcripts_file.name
+        self.config.args.out_dir = Path(config.args.out_dir) / self.run_id
+        self.config.args.tmp_dir = \
+            Path(config.args.tmp_dir) / f"tmp_{self.run_id}"
+        self.config.args.path_1_processed = self.config.args.path_1
+        self.config.args.path_2_processed = self.config.args.path_2
+        self.config.args.t_file_processed = (
+            config.args.tmp_dir / config.args.transcripts_file.stem
+            if config.args.transcripts_file.suffix == ".gz"
+            else config.args.tmp_dir / config.args.transcripts_file.name
```
uniqueg, line 64: Ah, this is probably what leads to the `E1101` issues? Seems to me that those should be added to the `Args()` model, maybe as `Optional[Path]` with `None` as default. It's semantically weird, because we are mixing non-CLI args with others, but I don't see a better solution right now (without bigger refactorings)

uniqueg (changes requested):
Really good work @BorisYourich, thanks a lot. Added a few comments. Perhaps not entirely necessary, but they would perhaps help to improve it somewhat

htsinfer/htsinfer.py
BorisYourich: This was very ugly I didn't realize it but removed the defaults and kept Path() instead, unfortunately it is not the place where the warning was coming from, the latest commit has the `# pylint: disable=E1101` comment as suggested in the previous comment. (In the htsinfer/cli.py line 141)

uniqueg (changes requested):
Changes look great - thanks a lot!
But can you please remove the `.idea` folder and contents (and maybe add `.idea/` to your global or project-specific Git ignore commit patterns)? I'll merge afterwards

uniqueg (commented):
Wrong PR, forget comment

---

Just saw that there are now 5 lines in `htsinfer/htsinfer.py` that are not covered in unit tests (when a `WorkEnvProblem` is raised. I suppose that this wasn't yet a thing before the last couple of merges.

Would you mind including a test for these before I merge, @BorisYourich?

PR #99, opened 2022-11-10
feat: infer presence of polyA tail fragments

### LINKED ISSUE #86: Infer presence of polyA tails in reads
TBA

TBA

TBA

**Additional context**
TBA

BorisYourich:
### Description

Added the cutadapt step in get_read_layout function. The report is parsed and the fraction of reads containing polyA tails is reported.

Also changed the __.readthedocs.yaml__ to use mamba instead of conda.

Fixes #86

### Type of change

      to not work as expected)
- [ ] Documentation update

### Checklist

_or do not apply_.

      warnings
      works
      by the proposed changes

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -2,6 +2,11 @@
 # Read the Docs configuration file
 # See https://docs.readthedocs.io/en/stable/config-file/v2.html for details
 version: 2
+
+build:
+  os: "ubuntu-20.04"
+  tools:
+    python: "mambaforge-4.10"
 sphinx:
    configuration: docs/api/conf.py
 formats:
--- a/README.md
+++ b/README.md
@@ -64,9 +64,11 @@ example library:
    "read_layout": {
       "file_1": {
          "adapt_3": "AATGATACGGCGACC"
+         "polyA_frac": 10.0
       },
       "file_2": {
          "adapt_3": "AATGATACGGCGACC"
+         "polyA_frac": 10.0
       }
    },
    "read_orientation": {
--- a/environment.yml
+++ b/environment.yml
@@ -13,5 +13,6 @@ dependencies:
   - pysam>=0.16.0
   - python>=3.6, <=3.10
   - star>=2.7.6
+  - cutadapt>=3.5
   - pip:
     - -e .
--- a/htsinfer/exceptions.py
+++ b/htsinfer/exceptions.py
@@ -40,3 +40,7 @@ class WorkEnvProblem(Exception):
 
 class TranscriptsFastaProblem(Exception):
     """Exception raised when an invalid transcripts fasta file is passed."""
+
+
+class CutadaptProblem(Exception):
+    """Exception raised when running cutadapt commands."""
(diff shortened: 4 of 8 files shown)


BorisYourich:
I added the polyA_frac to README output example which I forgot previously. I also made the changes in the Dockerfile. Previously the Image used conda to install dependencies, I removed that and built from python:3.9-slim. The image is 416MB compared to 1.77GB previously, builds in 30 seconds compared to 30 minutes of the previous one, and also alleviates some overhead as repeatedly the htsinfer called on the test files paired adapters finished half a second earlier than the previous image.
The Codecov report seems rather strange, as locally the line-rate is 1 on my computer, and the Codecov page also doesn't mention any files which should contain missed lines.

uniqueg (changes requested):
This is great!

Few minor comments:
- I strongly prefer to do one thing per feature branch/PR, because it makes the history cleaner, more precise and easily allows to roll back changes in case of issues. So I think the Dockerfile changes (which are great!) should really go in a different branch/PR, each with a properly phrased and formatted semantic/conventional commit message (https://www.conventionalcommits.org/en/v1.0.0/) as title (e.g., `build: Docker builds without Conda` and `feat: infer presence of polyA tail fragments`). Alternatively, you could use a single branch, but clean it up so that there is _exactly_ one clean commit per feature, again, using semantic commit messages. However, that is less preferred, because interactive rebasing of published branches is suboptimal - plus things get difficult if changes are requested (potenitally requires additional rebases).
- Don't think the CodeCov is your fault, it seems to have been there before. Perhaps you could open an issue for it so that we will get to the bottom of the problem and bring this back to 100% soon.

uniqueg:
There is also this issue for the Read the Docs build (I tried rerunning but repeatedly got the same result): 
image

Not sure if this is because of the addition of Cutadapt, but we didn't have this before. Not sure if this would potentially work with Mamba, but then I'm not sure how to tell the Read the Docs build system to use Mamba instead of Conda... Other issues may be that there isn't a good channel listed for Cutadapt, or at least not for the specified version. This underlying issue could also explain why the Conda-based Docker build took so extremely long. Anyway, something seems wrong here...

BorisYourich:
> There is also this issue for the Read the Docs build (I tried rerunning but repeatedly got the same result): image
> 
> Not sure if this is because of the addition of Cutadapt, but we didn't have this before. Not sure if this would potentially work with Mamba, but then I'm not sure how to tell the Read the Docs build system to use Mamba instead of Conda... Other issues may be that there isn't a good channel listed for Cutadapt, or at least not for the specified version. This underlying issue could also explain why the Conda-based Docker build took so extremely long. Anyway, something seems wrong here...

We might also try installing with pip, I didn't try in this commit but if it fails again I will push another one with cutadapt as pip dependency.

BorisYourich:
Seems that there is no advantage to using pip to install cutadapt, anyway I have found this: https://docs.readthedocs.io/en/stable/guides/conda.html
So now it uses mamba... let's see.

uniqueg:
Cool!

Guess I'll close this in favor of the two new PRs

PR #102, opened 2022-11-28
Infer polyA tail & Conda to Mamba

BorisYourich:
### Description

Fixes #86
Fixes #101 

### Type of change

Please delete options that are not relevant.

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality
      to not work as expected)
- [x] Documentation update

### Checklist

Please carefully read these items and tick them off if the statements are true
_or do not apply_.

- [x] I have performed a self-review of my own code
- [x] My code follows the existing coding style, lints and generates no new
      warnings
- [x] I have added type annotations to all function/method signatures, and I
      have added type annotations for any local variables that are non-trivial,
      potentially ambiguous or might otherwise benefit from explicit typing.
- [x] I have commented my code in hard-to-understand areas
- [x] I have added ["Google-style docstrings"] to all new modules, classes,
      methods/functions or updated previously existing ones
- [x] I have added tests that prove my fix is effective or that my feature
      works
- [x] New and existing unit tests pass locally with my changes and I have not
      reduced the code coverage relative to the previous state
- [x] I have updated any sections of the app's documentation that are affected
      by the proposed changes

If for some reason you are unable to tick off all boxes, please leave a
comment explaining the issue you are facing so that we can work on it
together.

PR #103, opened 2022-11-28
build: Docker builds without Conda

### LINKED ISSUE #96: Docker image for the htsinfer project
This Issue stems from the fact that __conda__ on macOS(Monterey and possibly higher) with arm64 chip is not able to create htsinfer environment due to the following error:
```
Solving environment: failed

ResolvePackageNotFound: 
  - kallisto[version='>=0.46.1']
  - star[version='>=2.7.6']
  - pysam[version='>=0.16.0']
```

I couldn't find any user-friendly fix or workaround, and since I sometimes develop on a mac I decided to create a docker image.

NA

**Additional context**
The image can be downloaded by ```docker pull edirexbot/htsinfer:dev``` and executed via: 
```
docker run --platform linux/amd64 \
      -v $HOME/htsinfer/tests/files/:/tmp \
      edirexbot/htsinfer:dev htsinfer /tmp/adapter_single.fastq
```
__--platform__ switch is there specifically for the __arm64__ architecture of the mac M1 chip and is not necessary. The volume is mounted on the ```/tmp``` directory in the container in this case, but similarly the command:
```
docker run edirexbot/htsinfer:dev htsinfer tests/files/adapter_single.fastq
```
could have been executed instead, as the entire ```htsinfer``` repo is inside the container in the ```/htsinfer``` directory.

The image is built by creating the __conda__ environment inside the image. This is just a temporary solution that was easy to implement. If the container version of ```htsinfer``` is something that is desired, I could install the dependencies without the conda and make the image a bit more lightweight and also include the docker image build in the CI/CD pipeline when the dev branch is changed.

BorisYourich:
### Description

Previously the Image used conda to install dependencies, now only apt-get and pip are used. The image is built from python:3.9-slim, which shaves off about 1.3GB in size. The image is 416MB compared to 1.77GB previously, builds in seconds compared to minutes, and also alleviates some overhead as repeatedly the htsinfer called on the test files paired adapters was half a second faster than the previous image.

Fixes #96 

This PR depends on #102 and should be merged after that one.

### Type of change

### Checklist

_or do not apply_.

      warnings
      works
      by the proposed changes

together.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,8 @@
 ###### BASE IMAGE ######
-FROM continuumio/miniconda3:4.11.0
+FROM python:3.9-slim
 
 ####### METADATA #######
-LABEL base_image="continuumio/miniconda3:4.11.0"
+LABEL base_image="python:3.9-slim"
 LABEL version="1.0"
 LABEL software="HTSinfer"
 LABEL software.version="v0.9.0"
@@ -21,24 +21,14 @@ LABEL maintainer.lab="Zavolan Lab"
 
 ##### INSTALLATION #####
 
-# COPY THE YAML & INSTALL SOFTWARE WITH CONDA
-WORKDIR /usr/src/app
+WORKDIR /htsinfer
+
 COPY ./ ./
-RUN conda env create --file environment.yml \
-    && conda clean --all
 
-# VARIABLES
-ARG WORKDIR="/home/USER"
-ARG USER="USER"
-ARG GROUP="GROUP"
-ENV PATH="${WORKDIR}:${PATH}"
+RUN apt-get update && apt-get install -y kallisto rna-star && rm -rf /var/lib/apt/lists/*
 
-# CREATE USER
-RUN groupadd -r ${GROUP} && useradd --no-log-init -r -g ${GROUP} ${USER}
+RUN pip install --upgrade pip && pip install biopython pandas pyahocorasick pydantic pysam cutadapt && \
+    pip install -e . && pip cache purge
 
-# SET ENVIRONMENT
-WORKDIR ${WORKDIR}
-RUN chown -R ${USER}:${GROUP} ${WORKDIR} && chmod 700 ${WORKDIR}
-USER ${USER}
-RUN echo "source activate htsinfer" > ~/.bashrc
-ENV PATH /opt/conda/envs/htsinfer/bin:$PATH
+ENTRYPOINT ["htsinfer"]
+CMD ["-h"]
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,12 +18,14 @@ jobs:
         uses: conda-incubator/setup-miniconda@v2
         with:
           auto-update-conda: true
+          mamba-version: "*"
+          channels: conda-forge,defaults
           environment-file: environment.yml
           activate-environment: htsinfer
           auto-activate-base: false
 
       - name: update env with dev packages
-        run: conda env update --file environment-dev.yml
+        run: mamba env update --file environment-dev.yml
 
       - name: display env info
         run: |
@@ -62,12 +64,14 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
           auto-update-conda: true
+          mamba-version: "*"
+          channels: conda-forge,defaults
           environment-file: environment.yml
           activate-environment: htsinfer
           auto-activate-base: false
 
       - name: update env with dev packages
-        run: conda env update --file environment-dev.yml
+        run: mamba env update --file environment-dev.yml
 
       - name: display env info
         run: |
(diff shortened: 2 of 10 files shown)


Dockerfile
```diff
@@ -21,24 +21,14 @@ LABEL maintainer.lab="Zavolan Lab"
...
-WORKDIR ${WORKDIR}
-RUN chown -R ${USER}:${GROUP} ${WORKDIR} && chmod 700 ${WORKDIR}
-USER ${USER}
-RUN echo "source activate htsinfer" > ~/.bashrc
-ENV PATH /opt/conda/envs/htsinfer/bin:$PATH
+
+RUN apt-get update && apt-get install -y kallisto rna-star && rm -rf /var/lib/apt/lists/*
+
+RUN pip install --upgrade pip && pip install biopython pandas pyahocorasick pydantic pysam cutadapt && \
+    pip install -e . && pip cache purge
```
uniqueg, line 31: I have two issues with this:

1. Without specifiying the version numbers, they are unlikely to be the same as when installing dependencies with Conda/Mamba, therefore we cannot guarantee the same behavior with the Dockerfile. Also, I'm not quite sure whether a given Ubuntu version will always install the same software versions via `apt` (I think so, but I'm not sure), so this may potentially not be reproducible.
2. Like this we need to list every dependency twice, once in `environment.yml` and once here. This is always a potential source of errors.

Both of these problems are particularly troublesome because we do not test the Docker image as extensively as the Conda/Mamba-based installation.

In fact, I tend to think that this is really a huge drawback of a non-Conda/Mamba-based image in general. It's gonna be hard to ensure that we have comparable environments in Docker and Conda/Mamba.

The other thing is that we are not planning to maintain the Dockerfile indefinitely. The plan is to publish HTSinfer on BioConda, which will automatically lead to a (Conda-based) container image being built for BioContainers. So I'm not sure the large image size is an issue for us now.

What do you think?

BorisYourich: Considering we want to keep the image slick, I could separate the pip dependencies into the `requirements.txt` file which would be used in the `Dockerfile` as well as in the `environment.yml`: 
```
pip:
  - -r requirements.txt
```
this way we would keep only the non-pip-installable dependencies in the `environment.yml`. 

The apt-get dependencies are a bit tricky as the version names from `apt` are not coherent with the `conda` naming eg. `kallisto=0.46.2+dfsg-3` vs. `kallisto==0.46.2`, but either we accept that there are two places where `Kallisto` and `STAR` package version is defined, or I write a bash one-liner that reads the environment.yml and downloads and builds from the source the mentioned version in the `environment.yml`. 

In any case, if there is a need for some other dependencies which cannot be installed with `pip` then the `Dockerfile` needs to be updated accordingly, but I don't think this is such a problem. The dependencies are not to be changed frequently anyway, and taking care of the docker image alongside the `conda` environment is not a big overhead, and the guidelines can be extended by a simple instruction such as: "If you add a dependency into the environment.yml file, include it also in the `Dockerfile`." 

But hey, If the plan was to build the biocontainers-based image from the beginning I didn't have to work on this one. I just think that having a `conda` environment inside the docker image is kind of useless since you can just install the conda environment and lose the overhead, unless there is a very specific use case, say Kubernetes cluster that does not allow users to install additional software, in that case having conda in the image/pod is pretty handy for installing extra packages in runtime.

uniqueg: Well, we need the Docker image almost exclusively because we want to run HTSinfer inside workflows. However, to ensure reproducibility, it's always best to use Biocontainers, even if they tend to be bigger. Images don't stay on Docker Hub forever, as recent times have shown, and managing this independently takes work that we might rather use elsewhere. The reason why we added the Dockerfile in the first places was to be able to use it in the ZARP/HTSinfer workflow, before it's available on BioConda (and hence Biocontainers).

I thought that you added a fix for the issue with M1/Arm because you needed that for something. Personally, I don't think there's a need to support a HTSinfer Docker image that runs on Mac at all (let alone M1). Docker is anwyay running on VMs in Mac, so I don't think that people running bioinfo tools in Mac is really such a common use case. As long as people can use the tool on Mac (via Conda) or in a workflow (via either Conda or Docker on Linux), I think we're good.

The decision for BioConda prepare images via Conda for Biocontainers is not to make the images as small as possible, but to ensure that there's an automatic pipeline that works in all cases. Again, this is a tradeoff where it was decided by BioConda/Biocontainers people that it's worth to take the extra size for maintenance reasons. And personally I think it's a reasonable decision.

If you don't want the work to be lost, I'm okay with merging it now (as is) and we could use it in workflows until it's available on Biocontainers. But after I would still remove it because I wouldn't want to either maintain or even promote an image if there's a corresponding one on Biocontainers, because in the community we _want_ people to use Biocontainers. If someone needs an optimized image because they keep reusing that image, they can build it themselves, but I wouldn't want people to raise issues on it, devs to remember to update it and deal with the issues raises above and alll of that.

BorisYourich: Sure, I completely understand, It is a reasonable tradeoff knowing how many hours I burned just trying to build the image from alpine which was a quite shortsighted decision, I can imagine what mess future changes in package versions might bring to this. 

The M1 issue was solely with conda because some packages are not available from any channels for the arm64, therefore I built the image and saw an opportunity to make it better, but yeah, it was the wrong direction of better in this case. However, I would be very glad If you merge this for now. The future Dockerfile will have to be different from this one and also from the one before so I will just sleep calmly knowing that it is stored somewhere.

uniqueg: Sure! And sorry about the delay, have really been flooded with work. Will be merging now

uniqueg (approved):

PR #124, opened 2023-04-27
perf: optimize read layout inference and defaults

### LINKED ISSUE #107: Analyze 3' end adapters
We will need the test suite to automatically generate a number of plots:

- A bar chart with the total number and the fraction of libraries for which an adapter was and was not identified
- A histogram of the fractions of reads that contained the most prevalent adapter fragments
- A histogram of the ratios of the fractions of the reads that contained the most prevalent vs the second most prevalent adapter frament

It should also generate a report with the actual numbers for the above (for the distributions, report some percentiles, e.g., 1, 2, 5, 10, 15, 25, 50, 75, 85, 90, 95, 98, 99).

The plots and report should inform the choice for default values for the following parameters in an informed manner:
- `--read-layout-min-match-percentage`
- `--read-layout-min-frequency-ratio`

For example, we could set the values such that, say, 95% of samples would have values above the defaults. But the best is to look at the actual histograms and see if they suggest a natural cutoff.

After running HTSinfer on all test samples, at least for some samples, no adapter will be identified.

For libraries for which there are no adapters identified (all zero), even after adding additional adapters in #106, we can try the following approach to see if they perhaps do contain an adapter that is not in our list:

- Prepare a truncated version of the library that only contains the last $k$ nucleotides
- For the chosen $k$, construct all (i.e., $4^k$) possible $k$-mers 
- For each $k$-mer, count the number of occurences in the truncated library 
- Inspect the distribution of counts: is there a single high entropy outlier (not something like `AAAAAA`) that is significantly more frequent than other $k$-mers?
- If so, try identifying a full 15 nt adapter candidate sequence by searching for the $k$-mer in the original library and comparing what's up-/downstream of each match (can be done automated, but probably easier to just do manually)
- Search for candidate in _other_ libraries: in those libraries where an adapter was already clearly identified, there should be (almost) no counts for the new adapter; but in other libraries for which no adapter was found, there may perhaps be many counts
- Depending on the outcome of the search, include the adapter candidate (15 nts) 

> A good range for $k$ might be between 5 and 8; the longer the $k$, the more clear will be the result; however, also the counts per $k$ will be lower and the counting process will take longer (exponentially)

balajtimate:
### Description

Based on the results from the testing data, the default parameter was optimized:
- Shortened adapter fragments to a length of 12, removed duplicates
- Updated the default parameter for `--read-layout-min-match-percentage` to 0.1
- Updated `README.md`

Fixes #107

### Type of change

- [X] Documentation update

### Checklist

_or do not apply_.

      warnings
      works
      by the proposed changes

together.


uniqueg (approved):
Fantastic - thanks so much :)

Guess at some point we could look into removing some of the places where defaults are defined or described, e.g., in `README.md`). Possibly there is also a way of reusing Pydantic models for the CLI argument definition. But that's just FYI and not urgent.

PR #125, opened 2023-05-03
perf: optimize STAR memory usage and read orientation defaults 

### LINKED ISSUE #115: STAR fails to create index
This might not be directly related to HTSinfer itself, but when running HTSinfer on my PC with the samples [SRR19067720](https://www.ncbi.nlm.nih.gov/sra/?term=SRR19067720) and [SRR12414066](https://www.ncbi.nlm.nih.gov/sra/?term=SRR12414066), the process terminates and I get the error message 'Failed to create STAR index'. Both samples are from the same organism Pongo abelii (even though the size of the samples are very different, 6.5 GB vs 166 MB), which I suppose is not a frequent model organism, but it also happened with a human sample too, when I used the `--records` set to default. 

It's probably a memory issue, because the termination happens after throwing an instance of `std::bad_alloc`; also, it doesn't seem to happen when running HTSinfer on the cluster.

I'm just brining this up because it could potentially arise when other end-users run it on their own setup (and not on a cluster).
Maybe we could just add a minimum memory recommendation in the readme?

balajtimate:
### Description

- Added `--genomeChrBinNbits` to `get_read_orientation.py` to calculate the size of bins for genome storage as `min(18,log2[max(GenomeLength/NumberOfReferences,ReadLength)])`
- Updated the default parameter for `--read_orientation_min_mapped_reads` to 18
- Updated `README.md`

Fixes #115 

### Type of change

- [X] Documentation update

### Checklist

_or do not apply_.

      warnings
      works
      by the proposed changes

together.


README.md
```diff
@@ -163,7 +163,7 @@ optional arguments:
                         considered the library's source.
   --read-orientation-min-mapped-reads INT
                         minimum number of mapped reads for deeming the read
-                        orientation result reliable (default: 20)
+                        orientation result reliable (default: 18)
```
uniqueg: Hmmm, I think 18 is so random that it gives a suggestion that we somehow determined this to be optimal in some way. I'd rather leave the default as it was and modify the test file such that it will have alignments for all 20 reads. Or alternatively set the param to 10 for the failing test.

uniqueg (changes requested):
I think it's just this one issue, rest looks good to me

README.md
balajtimate: Yeah you're right, sorry, I changed back the default to 20 and updated the tests with the parameter.

uniqueg (approved):

PR #129, opened 2023-05-24
feat: infer lib type if mate info not in seq IDs

### LINKED ISSUE #85: Infer library type if sequence identifiers do not contain mate info

When depositing FASTQ files to the [Sequence Read Archive (SRA)](https://www.ncbi.nlm.nih.gov/sra), the original sequence identifiers are not preserved. In the sequence identifiers generated by the SRA, mate information is not retained. Therefore, the current regex-based inference of the library type (single vs paired) does not work for FASTQ samples derived from the SRA.

As using HTSinfer on libraries obtained from SRA is an important use case, this is a critical problem.

Whether two different libraries are mates of one another can also be inferred indirectly, based on two assumptions:
1. Mates have the same sequence identifiers
2. Mates of a paired-ended library generally (though possibly not always) map to the same transcripts. Also, the alignment orientation of each mate should match the ones expected based on the read orientation.

So, for libraries whose type cannot be inferred based on the regex approach (faster!), mate library files could be individually mapped to the set of transcripts (or mappings used from another step) and the alignments could be analyzed for the fraction of mate pairs (reads with same identifier) that map to the same transcript.

Alternatively, one could possibly use a mapping strategy where both mate library files are supplied.

BorisYourich:
### Description

I have added a GetOrientation class instance in the GetLibType class so that when
the sequence identifiers do not match, the files will be mapped separately. 
Then the alignments for each file are compared and if at least a cutoff fraction of the reads
can be considered concordant, the results.relationship is changed to "split_mates".

Note that this implementation is completely written in Python and will easily become the bottleneck
of the program as the logic by which the mappings are compared is all-vs-all for the two mates, so
if a high fraction of reads are multi-mapping, the execution will get significantly slower

Fixes #85 

### Type of change

### Checklist

_or do not apply_.

      warnings
      works
      by the proposed changes

together.


htsinfer/cli.py
```diff
@@ -212,6 +212,29 @@ def __call__(
...
+    parser.add_argument(
+        "--library-type-mates-cutoff",
+        dest="lib_type_mates_cutoff",
+        metavar="FLOAT",
+        type=float,
+        default=0.95,
+        help=(
+            "minimum fraction of mates that can be mapped to compatible loci "
+            "and are considered concordant pairs / all mates. "
+        )
```
uniqueg: Maybe add here that these are only used when the lib type cannot be inferred by regex from the seq identifiers

README.md
```diff
@@ -154,13 +156,21 @@ optional arguments:
...
                         most frequent source in order for the former to be
                         considered the library's source.
+  --library-type-max-distance INT
+                        upper limit on the difference in the reference sequence
+			coordinates between the two mates to be considered as
+			coming from a single fragment. (default: 1000)
+  --library-type-mates-cutoff FLOAT
+                        minimum fraction of mates that can be mapped to compatible
+			loci and are considered concordant pairs / all mates. 
+			(default: 0.95)
```
uniqueg: Formatting seeems a little different than some other options (position and overall presence of `.` and the indentation.

However, I don't really care, as I think this should be removed from the `README.md` anyways. Not good to have this in too many places (docs, CLI args, models)...

uniqueg (changes requested):
This looks pretty good to me, at first glance. I can't review in depth as it's getting late and I have an early and long flight tomorrow.

@balajtimate: could you help review and test this, please?

balajtimate:
Sure, I'll run some tests and get back to it soon

balajtimate (commented):
I ran the tests on our sample list of 385 PE samples, and when changing the seq IDs so that the pairs don't match, this seemed to correctly identify the "split_mates" relationships. 

However, when running it on some unrelated samples, and the lib type cannot be determined (both are NoneType), I seem to get a TypeError in the `_compare_alignments` function in `get_library_type.py`. I think the error occurs during the comparision of the separate mappings. You could recreate it by running HTSinfer on the `unmapped_paired_mate_1.fastq` and `second_mate.fastq` test files. Could you look into this?

BorisYourich:
Sure, thanks for taking the time, I will fix that soon as possible.

BorisYourich:
The issue was that I didn't consider reads that do not map at all, there is no need to include such alignments, therefore I added a check.

balajtimate (dismissed):
Perfect, it works great, thank you! I think if you make the changes Alex requested, it can be merged.

BorisYourich:
Okay, I just pushed the changes that Alex recommended, I think you can merge this.

uniqueg (approved):
Awesome job - thanks @BorisYourich and @balajtimate

PR #131, opened 2023-07-18
refactor: avoid duplicate mappings

### LINKED ISSUE #130: Create Mapping class for managing STAR cmd and outputs
Currently, the library type inference module uses STAR to analyze whether two files are mates or not if the sequence identifiers differ.
This way mapping will happen twice for the mates, having a standalone class that takes care of all STAR related data and information,
we can easily use an instance of this class in the program to check whether mapping has been performed and where the results are stored.

Define a "Mapping" class in a separate module `mapping.py`

Defining he class in the `utils.py` module

BorisYourich:
### Description

Part of the code from the `get_read_orientation.py` module was transferred into a new module `mapping.py` which takes care of the STAR mapping process. The instance of the `Mapping` class is created in the `HtsInfer` class and supplemented as an argument to the `get_lib_type` and `get_orientation` modules, so as to avoid a repeated mapping if the sequence IDs are not the same and the mates are actually not mates.

Fixes #130 

### Type of change

### Checklist

_or do not apply_.

      warnings
      works
      by the proposed changes

One Issue which I was not able to replicate is the unit test failures in the CI workflow, I will not be able to work on this currently, so I leave it as is for now.


htsinfer/cli.py
```diff
@@ -212,6 +212,31 @@ def __call__(
...
+        "--library-type-max-distance",
+        dest="lib_type_max_distance",
+        metavar="INT",
+        type=int,
+        default=1000,
+        help=(
+            "upper limit on the difference in the reference "
+            "sequence coordinates between the two mates to be "
+            "considered as coming from a single fragment "
+            "(Used only when sequence identifiers do not match) "
```
uniqueg: For consistency I would use all lowercase "used". Also in the next option.

htsinfer/get_library_type.py
```diff
@@ -116,6 +124,111 @@ def _evaluate_mate_relationship(
                 self.results.relationship = (
                     StatesTypeRelationship.split_mates
                 )
+                self.mapping.library_type.relationship \
```
uniqueg: Better to use parentheses for wrapping lines. See PEP 8 style guide.

htsinfer/get_read_orientation.py
```diff
@@ -416,6 +101,7 @@ def process_alignments(
             if star_dirs[0].name == "file_1":
                 results.file_1 = self.process_single(sam=paths[0])
         elif len(star_dirs) == 2:
+            print("Getting there")
```
uniqueg: I think this is a leftover from testing. If not, please change into a logging message

uniqueg (changes requested):
Good stuff, thanks a lot. Please check the unit tests and resolve issues. Otherwise only minor comments from my side. Once the tests pass, can you please tag @balajtimate for an additional pair of eyes? 🙏

htsinfer/get_read_orientation.py
BorisYourich: opsie 😆

uniqueg:
@BorisYourich: I've been away for a while. Should I look at this again or was there nothing added/changed since the last review?

BorisYourich:
> @BorisYourich: I've been away for a while. Should I look at this again or was there nothing added/changed since the last review?

Hey, sorry for such a late response @uniqueg I was busy with other stuff, no update on this yet.

uniqueg:
No problem. Thanks for the update on no update :) I had just asked because I was away for quite a while myself and lost a bit track of progress on different ends

BorisYourich:
I made the required changes plus something extra I found, but the unit tests still fail and I have no idea why, and it seems random, take a look at the replayed 3.7 unit test, many more tests failed compared to previous run. I cannot replicate in my local env, even if I downgrade python to 3.7, the tests still pass.

uniqueg:
Hi @BorisYourich, I've fixed some of the errors by stabilizing the build in various ways (remote now installs the exact same package as local for me): https://github.com/zavolanlab/htsinfer/pull/131/files/0a232822922a006fd97ee9d6e94d44db733eca5d..b5bd469df2694f20ba46178425752b036cda2f94

Among other things, I have included Py3.10 in the supported versions, but dropped support for Py3.7 - support for which is phasing out anyway.

As you can see, there is still one issue remaining that affects two tests. These two tests also pass on my local machine (I suppose the same issue you faced), which -together with the error message- makes me think that it's a permission problem. Maybe a file is attempted to be written in a location that we do have write access on our local machine, but wouldn't have write access on the remote machine. Or a file is created in a temporary directory that is garbage collected on the remote between tests, but not locally.

From the error stack, I would have a look here first:
https://github.com/zavolanlab/htsinfer/blob/b5bd469df2694f20ba46178425752b036cda2f94/htsinfer/mapping.py#L79-L142

This method also had a line that created a file name based off an object (which I first thought was the problem). I have now given the transcripts subset file a hardcoded name (not sure if that could cause problems in some case?). And still this stringified library source object instance is used in the log message, so this isn't handled well. Seems to me that this function needs guarding against edge cases, e.g., when there is no library source determined/available).

As I think you know that particular code better from your changes, I would like to pass it back to you if possible. With some debug messages and maybe additional try-except blocks around the code reading and writing that subset transcripts file, I'm sure you will find out what the problem is on the remote.

BorisYourich:
Playing around with the tests I realized where the problem is, the tests didn't have the `tmpdir` argument. Everything works now :)

uniqueg:
Awesome, thanks a lot, @BorisYourich!

@balajtimate: Can you please test the feature branch to see if the proposed changes break anything on real world data? Maybe you can run it on a few samples and see if you get the same results as you did previously.

If all is good, I will merge.

balajtimate:
Okay, so I tested on a subset of 20 samples from our sample data, random organisms, SE/PE and there was no error, code ran perfectly, and the results match the ones from the run with the dev branch. 😄 

I think you can merge now, if the 3.7 integration and unit tests run.

uniqueg:
Thanks! Merging this now, given that it doesn't break anything. We can still do in-depth testing later to see if it actually works as expected. Thanks a lot @BorisYourich

PR #133, opened 2023-08-21
feat: infer additional statistics from read lengths

### LINKED ISSUE #132: Expand inferred read length stats
In many cases, the reported read length in SRA does not explain the inferred read length results (e.g. [SRR22397141](https://trace.ncbi.nlm.nih.gov/Traces/?view=run_browser&acc=SRR22397141&display=metadata), reported: 147 bp, inferred min: 35, max: 151). In PE samples, the reported length sometimes can be the sum of the maximum inferred length of forward and backward reads, but not always ([SRR19776630](https://trace.ncbi.nlm.nih.gov/Traces/?view=run_browser&acc=SRR19776630&display=metadata), reported: 585 (292 + 293), inferred min: 35, max 301 for both reads.)
In general, it would also be beneficial to get more info about the distribution of the read lengths in the sample (e.g. if the median is closer to the min, that could also raise some red flags in regards to the quality of the sequencing etc)

Add mean, median and mode of the read lengths to the reported library stats.

balajtimate:
### Description

- Rewrite `get_library_stats.py` to infer mean, median and mode from the sample read length
- Add additional adapter fragment sequence
- Refactor `htsinfer.py` so the results print to STDOUT is now sorted in order of inference (instead of alphabetical) 
- Update logging messages and `README.md`
- Downgrade Kallisto version as it was causing issues with CI

Fixes #132 

### Type of change

### Checklist

      warnings
      works
      by the proposed changes


environment.yml
```diff
@@ -5,7 +5,7 @@ channels:
   - conda-forge
 dependencies:
   - biopython>=1.78
-  - kallisto>=0.46.1
+  - kallisto=0.48.0
```
uniqueg: Still better to use a range. For example, >=0.46.1, <= 0.48.0

htsinfer/get_library_stats.py
```diff
@@ -18,8 +20,8 @@
 
 
 class GetLibStats:
-    """Determine library statitics of a single- or paired-end seguencing
-    library.
+    """Determine library statitics of a single- or paired-end \
+        sequencing library.
```
uniqueg: Check PEP 8 and the Python style guide on docstrings (I think PEP 257)  to see how to format docstrings. Basically, if not everything fits on one line, the first line should contain only `"""`. Also, there should definitely not be any backslashes inside for wrapping lines.

htsinfer/get_library_stats.py
```diff
@@ -48,21 +51,31 @@ def evaluate(self) -> ResultsStats:
         stats = ResultsStats()
 
         # process file 1
-        LOGGER.info(f"Obtaining statistics for file: {self.paths[0]}")
-        stats.file_1.read_length.min, stats.file_1.read_length.max = (
-            self.fastq_get_min_max_read_length(fastq=self.paths[0])
+        LOGGER.info("Obtaining statistics for file: %s", self.paths[0])
```
uniqueg: Better to be consistent and use an f-string. Also for other log messages below

htsinfer/htsinfer.py
```diff
@@ -310,6 +310,6 @@ def print(self):
         sys.stdout.write(
             self.config.results.json(
                 indent=3,
-                sort_keys=True,
+                sort_keys=False,
```
uniqueg, line 313: Why did you do that?

htsinfer/models.py
```diff
@@ -81,23 +81,34 @@ class ReadLength(BaseModel):
...
+    median: Optional[float] = None
+    mode: Optional[int] = None
 
 
 class Stats(BaseModel):
     """Library statistics of an individual sequencing file.
 
     Args:
-        read_length: Tuple of minimum and maximum length of reads in library.
+        read_length: Tuple of mininimum, maximum, mean, median and mode \
```
uniqueg: Again, please do not use `\` in docstrings

uniqueg (changes requested):

htsinfer/htsinfer.py
balajtimate: With the added read length stats, the sorted keys get reported alphabetically as: "max, mean, median, min, mode" which just looks weird to me, as the order they inferred is "min, max, mean, median, mode" :sweat_smile: Should I change it back?

uniqueg: Is there some consistency in the sort order for subproperties for other inferences (children, like here)? And for the different inferences themselves (parents)? If so, we should stick to it, I think, regardless of how it looks. Principle of least surprise ;-)

balajtimate: Actually there's none. When unsorted, the parent keys just get reported in the order they're inferred (which I kinda think makes more sense): `library_stats` (read length stats), `library_type` (relationship between read pairs), `library_source `(organism), `read_orientation`, `read_layout`, while when sorted, it is: `library_source`, `library_stats`, `library_type`, `read_layout`, `read_orientation`. Apart from the stats mentioned above (min, max etc), the children are the same. So I suppose if you deem the inferred organism the most important and put it first, it makes sense to sort it.

uniqueg: Well, if there's no consistent sort order, then just leave it "unsorted" :)

environment.yml
balajtimate: Thanks, fixed in commit.

htsinfer/get_library_stats.py
balajtimate: Fixed in commit.

htsinfer/models.py
balajtimate: Fixed in commit.

htsinfer/get_library_stats.py
balajtimate: Thanks, fixed it.

uniqueg (approved):

PR #146, opened 2023-11-09
fix: update Pydantic version

balajtimate:
### Description

- Update range of supported Pydantic versions (>=2, <3)
- Replace `json` method with `model_dump_json`, remove `dumps_kwargs` argument
- Convert tmpdir path to str before passing as arg by adding `.strpath` attribute to tmpdir in `test_htsinfer.py`, as pydantic was not recognizing at as a valid path

### Type of change

### Checklist

_or do not apply_.

      warnings
      works
      by the proposed changes

together.

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -13,4 +13,4 @@ formats:
    - pdf
    - epub
 conda:
-  environment: environment.yml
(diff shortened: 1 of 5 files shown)


uniqueg (approved):
Well done :)

PR #147, opened 2023-11-10
feat: add tax_id parameter

balajtimate:
### Description

To have better control over the read orientation inference process, allow users to define the organism of the given sample, instead of inferring it. When the organism ID is given, the transcripts file is filtered based on that organism, and the filtered file is used for to infer the orientation.

- Add parameter `--org-id` (int) to specify the taxonomy ID of the organism to map the reads against
- Add function to find the short name belonging to the given org-id, to be reported in the results; if org-id is not found in transcript list, infer it
- Add tests
- Update `README.md` and `.gitignore`
- Fix typo in `get_read_layout.py` and pylint config, remove unused option value 'C0330'
- Replace `.json` method with `.model_dump_json` for Pydantic version >2 in `htsinfer.py` (in connection to #146)

### Type of change

### Checklist

_or do not apply_.

      warnings
      works
      by the proposed changes

together.


.gitignore
```diff
@@ -117,3 +117,4 @@ tests/.DS_Store
 results_htsinfer
 .snakemake
 tests/cluster_tests/results_sra_downloads
+*.out
```
uniqueg, line 120: This file shouldn't be under version control. Please remove it and consider adding `.DS_Store` to `.gitignore`

README.md
```diff
@@ -109,6 +109,7 @@ htsinfer [--output-directory PATH]
          [--library-type-mates-cutoff FLOAT]
          [--read-orientation-min-mapped-reads INT]
          [--read-orientation-min-fraction FLOAT]
+         [--org-id INT]
```
uniqueg: Probably less ambiguous to name the param `--tax-id`

htsinfer/cli.py
```diff
@@ -262,6 +262,17 @@ def __call__(
...
         )
     )
+    parser.add_argument(
+        '--org-id',
+        dest="org_id",
+        metavar="INT",
+        type=int,
+        default=None,
+        help=(
+            "source organism of the sequencing library; if provided, will not "
```
uniqueg: Consistent with the comment above, it'd be important to clarify what type of identifier is expected here (I think it would "NCBI taxonomy identifier", but please double check if that is the "official" name)

htsinfer/get_library_source.py
```diff
@@ -5,6 +5,8 @@
 import subprocess as sp
 import tempfile
 
+from typing import Optional
```
uniqueg: The `typing` module is built in, and imports from built-in modules should be in a different block than those from third party modules/packages.

htsinfer/get_library_source.py
```diff
@@ -71,16 +75,59 @@ def evaluate(self) -> ResultsSource:
...
+            )
+
+            if org_name is not None:
+                source.file_1.short_name = org_name
+
+                if self.paths[1] is not None:
+                    source.file_2.taxon_id = self.org_id
+                    source.file_2.short_name = source.file_1.short_name
+
+            else:
```
uniqueg: I think this behavior may violate the principle of least surprise: If the user supplied the sample source, it is reasonable to assume that they are sure about the sample source. And so they would be surprised to find that eventually we ended up overriding their conviction.

Rather, I think if we do not find the supplied taxonomy ID in our list of supported sample source (and hence found no corresponding short name), we should stop the program with an explicit error message telling the user that the provided taxonomy ID is not supported.

Note that the case where we _do_ find a source and the source is supported but we end up without any transcripts (regardless of whether it was inferred or supplied by the user), as mentioned in the PR description is a different case: It's an error on _our_ side, because it would indicate that we have included a source in the list of supported sources that is in actual fact _not_ supported. This is one of the "shouldn't happen" errors that you could choose to catch (and exit gracefully with an "internal error", some detailed message on what happened and a request to the user to raise an issue to please report the error). Or you could choose not to explicitly catch (there will be many cases of things that "shouldn't happen" that we didn't think about, and we could just wait for users to report them if they do indeed happen).

Perhaps the better strategy for this last category of errors would be to define a generic "HTSInferException" that (all other custom exceptions inherit from) and that is raised at the highest level whenever there is an error raised that is otherwise uncaught. We could then use this last resort "trap" to ask users to file a bug report, including the traceback, on the HTSinfer issue tracker, together with a link).

But to implement that would of course be another issue.

uniqueg: Okay, I've just realized that we do not have a dedicated list of supported sources (I was confused because we have one in ZARP-cli) and that instead the transcripts file is our single source of truth for supported sources (which is fine).

This of course invalidates some of my previous comment. In particular, there is no difference between the cases where the user provided taxon ID is not supported and the situation where no transcripts are available - they are one and the same. And so we should treat this as a client error and let the user know that they provided an unsupported ID as described above.

The last part of the comment above might still be a useful improvement though, for errors on our side that we need the users help with to encounter and fix. So maybe you can create an issue for that.

htsinfer/get_library_source.py
```diff
@@ -71,16 +75,59 @@ def evaluate(self) -> ResultsSource:
...
         source = ResultsSource()
-        index = self.create_kallisto_index()
-        source.file_1 = self.get_source(
-            fastq=self.paths[0],
-            index=index,
-        )
-        if self.paths[1] is not None:
-            source.file_2 = self.get_source(
-                fastq=self.paths[1],
+        # Check if library_source is provided, otherwise infer it
```
uniqueg, line 78: There are a lot of conditionals in this function now. I would probably refactor `.get_organism()` such that it raises a custom `UnsupportedSampleSourceException` and then just drop the check that ensures that the organism name is set.

Btw, I think we have tried (or at least that was my intention) to avoid "organism" in favor of the more generic "source", because in principle (even if don't currently support it), a sample could be from different cell types/lines of the same organism.

So for consistency, it may be better to use `source` or `src` instead of `organism` or `org` in your code.

htsinfer/get_library_source.py
```diff
@@ -281,3 +328,45 @@ def get_source_expression(
...
+
+        Args:
+            taxon_id: Taxonomy ID of a given organism (int).
+            transcripts_file: Path to fasta file containing transcripts.
+
+        Returns:
+            Short name of the organism belonging to the given tax ID.
+
+        Raises:
+            Could not process input FASTA file.
```
uniqueg: In the `Raises` section, one should enumerate the different scenarios that can occur, starting with the exception that is raised. So, in case, start off with "FileProblem" (check the other docstrings for details on the formatting).

htsinfer/get_library_source.py
```diff
@@ -281,3 +328,45 @@ def get_source_expression(
...
+    @staticmethod
+    def get_organism_name(
+        taxon_id: int,
+        transcripts_file: Path,
+    ) -> Optional[str]:
+        """Return name of the organism, based on tax ID.
+
+        Args:
+            taxon_id: Taxonomy ID of a given organism (int).
+            transcripts_file: Path to fasta file containing transcripts.
```
uniqueg: "FASTA" in all caps, for consistency

htsinfer/get_library_source.py
```diff
@@ -281,3 +328,45 @@ def get_source_expression(
...
+
+    @staticmethod
+    def get_organism_name(
+        taxon_id: int,
+        transcripts_file: Path,
+    ) -> Optional[str]:
+        """Return name of the organism, based on tax ID.
+
+        Args:
+            taxon_id: Taxonomy ID of a given organism (int).
```
uniqueg: Don't annotate the type twice. It's included as a hint in the signature (the better place), so no need to repeat it here. If the API ever changes, one may forget to update it, leading to inconsistencies/confusion.

htsinfer/get_library_source.py
```diff
@@ -281,3 +328,45 @@ def get_source_expression(
 
         # return as dictionary
         return dat_agg.sort_values(["tpm"], ascending=False)
+
+    @staticmethod
+    def get_organism_name(
```
uniqueg: See comment about use of "organism" above

htsinfer/get_library_source.py
```diff
@@ -281,3 +328,45 @@ def get_source_expression(
...
+                    handle=transcripts_file,
+                    format='fasta',
+            ):
+                org_id = int(record.description.split("|")[4])
+                org_name = record.description.split("|")[3]
+
+                if org_id not in org_dict:
+                    org_dict[org_id] = org_name
+                else:
+                    org_dict[org_id] = org_name
```
uniqueg: So you are doing the same thing, regardless of whether you have already encountered this tax ID or not. Why the conditional then?

But this actually raises a concern: _Are_ there one-to-many relationships in our data with respect to taxon IDs? If so, we would need a way how to deal with them, e.g., remove these cases (and error out to guard against inadvertent reintroductions), merging them, supporting lists of short names...

What do you think?

htsinfer/get_library_source.py
```diff
@@ -281,3 +328,45 @@ def get_source_expression(
 
         # return as dictionary
         return dat_agg.sort_values(["tpm"], ascending=False)
+
+    @staticmethod
+    def get_organism_name(
+        taxon_id: int,
+        transcripts_file: Path,
+    ) -> Optional[str]:
```
uniqueg: Don't forget to change the return type if you refactor to raise an exception if the taxon ID wasn't found

htsinfer/get_library_source.py
```diff
@@ -281,3 +328,45 @@ def get_source_expression(
...
+                    org_dict[org_id] = org_name
+                else:
+                    org_dict[org_id] = org_name
+
+        except OSError as exc:
+            raise FileProblem(
+                f"Could not process file '{transcripts_file}'"
+            ) from exc
+
+        if taxon_id in org_dict:
```
uniqueg: When raising an exception instead, you can just return the organism/source name without the `if`, catching the `KeyError` instead and reraising it as the new custom exception for this particular case (see comment above).

htsinfer/htsinfer.py
```diff
@@ -85,15 +85,15 @@ def evaluate(self):
                 self.get_library_stats()
                 LOGGER.info(
                     "Library stats determined: "
-                    f"{self.config.results.library_stats.json()}"
+                    f"{self.config.results.library_stats.model_dump_json()}"
```
uniqueg, line 88: These changes should not show up here. Please merge the updated default branch that already has these changes from the other PR into your feature branch.

uniqueg (changes requested):

README.md
balajtimate: Fixed in commit

htsinfer/cli.py
balajtimate: Added "NCBI taxonomic identifier" to description

htsinfer/get_library_source.py
balajtimate: Fixed in commit :sweat_smile:

.gitignore
balajtimate: The `Log.out` is just the log from STAR, shouldn't that be ignored?
I add `.DS_Store` to the list.

htsinfer/get_library_source.py
balajtimate: Fixed in commit

htsinfer/get_library_source.py
balajtimate: Fixed in commit

htsinfer/get_library_source.py
balajtimate: So in a previous version, I implemented what you mentioned (stop the run when the tax ID is invalid or not found in transcript list), then changed it to just a warning, as my thinking was that the program should return _some_ result when it's run, but then the invalid tax ID with no corresponding source name would get reported back as the "inferred" source at the end of the run, and that didn't seem right. 

It's true that there is no difference in terms of results between the run with not supported tax ID and when inferring but no organism is found in the transcript list, but maybe there could be cases when the inferred org was actually what the user wanted in the first place, so inferring it despite given a wrong tax ID could be beneficial? 

For example, in our transcript list we have "Oryza sativa" with tax ID 39947, but when looking it up in the NCBI db, it turns out it's actually [Oryza sativa Japonica Group](https://www.ncbi.nlm.nih.gov/Taxonomy/Browser/wwwtax.cgi?mode=Info&id=39947&lvl=3&lin=f&keep=1&srchmode=1&unlock), and when searching for the tax ID for just "Oryza sativa", it's actually [4530](https://www.ncbi.nlm.nih.gov/Taxonomy/Browser/wwwtax.cgi?mode=Info&id=4530&lvl=3&lin=f&keep=1&srchmode=1&unlock). I know this is a very niche case, or maybe just a quirk from how the transcript list was collected from different subspecies and whatnot, so to me it made sense to report back a result with an inferred source, even when a wrong tax ID was an input. 

On second thought though, this could also backfire, if the inferred org is something totally different from the given tax ID (like giving an ID for some mouse subspecies that's not in our list, but a butterfly gets inferred), so throwing an error would be probably the best 😅 I'll refactor it.

.gitignore
uniqueg: Alright, never mind, thought somehow that there is a file `tests/.DS_Store` in the repo. But it's actually `.gitignore`, and `tests/.DS_Store` was a line in that file.

htsinfer/htsinfer.py
balajtimate: Left it here after all, this is different from the methods refactored in #146

htsinfer/get_library_source.py
uniqueg: Yes, there are some issues from the sourcing of the transcripts for the different species. It was, in fact, quite painful to prepare the list of source names and tax IDs, because there are actually quite a lot of inconsistencies.

Still, I think it's safer not to just do something that the user didn't ask for. If you supply the source via the tax ID, and we _know_ the tax ID, then we assume the user is correct - even if it's incorrect. Not our problem. And if we don't know the tax ID, we just tell them, even if they provided a tax ID for a source that actually _is_ supported - but under a different tax ID. If anything, we should change this by working on the sourcing of the genome resources, _not_ by adapting our program logic to it.

Now, if you are concerned about the results _suggesting_ that we inferred something that was actually given as an input, you could add a field to the results that mentions that the source ID was given as an input. Or better yet, a field that lists _all_ of the result fields that were given as an input (e.g., the tax ID), or in another way trivially derived from an input (e.g., the source name when the tax ID is given) - in case we also later allow users to provide other inputs to avoid inference.

htsinfer/get_library_source.py
balajtimate: Added the exception and refactored to use `source` instead.

htsinfer/get_library_source.py
balajtimate: Refactored it

htsinfer/get_library_source.py
balajtimate: Refactored it, now it throws an error if the taxon ID is not supported. 
I'll open a separate issue to return the results which were given as input.

htsinfer/get_library_source.py
balajtimate: Refactored it so that instead of iterating over each entry and checking if it's already in the dict, now just selects the unique key - value pairs from the list. 

For the one-to-many (or many-to-many) relationships between the tax-ids and organisms, I'm not sure, but I'd say right now, based on the transcript list we're building the tax-id: organism name dictionary from, there are no such cases where one tax id is matched to multiple organisms. If in the future, additional organisms are added to the list, this could be a problem however.

htsinfer/get_library_source.py
```diff
@@ -82,34 +82,11 @@ def evaluate(self) -> ResultsSource:
...
-                    source.file_2.short_name = source.file_1.short_name
-
-            else:
-                LOGGER.warning(
-                    f"Taxon ID '{self.tax_id}' not found in "
-                    "organism dictionary, inferring source organism..."
-                )
-                index = self.create_kallisto_index()
-                library_source = self.get_source(
-                    fastq=self.paths[0],
```
uniqueg: But are you even still calling `.get_source_name()` now? I don't see it anywhere

htsinfer/get_library_source.py
```diff
@@ -71,16 +75,36 @@ def evaluate(self) -> ResultsSource:
...
-            fastq=self.paths[0],
-            index=index,
-        )
-        if self.paths[1] is not None:
-            source.file_2 = self.get_source(
-                fastq=self.paths[1],
+        # Check if library_source is provided, otherwise infer it
+        if self.tax_id is not None:
+            source.file_1.taxon_id = self.tax_id
+            src_name = self.get_source_name(
```
balajtimate, line 81: Yeah, it's here, I just removed the check when the source name is not set @uniqueg

htsinfer/get_library_source.py
balajtimate: Please see my other comment. If tax-id parameter is given, it's assigned as the `source.file_1.taxon_id` and the `.get_source_name()` is called to get `source.file_1.short_name` (and same for `source.file_2`). If tax-id is not set, then both params are inferred.

htsinfer/exceptions.py
```diff
@@ -44,3 +44,9 @@ class TranscriptsFastaProblem(Exception):
 
 class CutadaptProblem(Exception):
     """Exception raised when running cutadapt commands."""
+
+
+class UnsupportedSampleSourceException(Exception):
+    """Exception raised when taxonomy ID is not found in the source
+    organism list.
```
uniqueg: According to [PEP 8](https://peps.python.org/pep-0008/) and [PEP 257](https://peps.python.org/pep-0257/), docstrings should either fit on one line:

```py
"""One-line docstring."""
```

OR they should be broken apart into a summary line and an extended description (also, for any multiline docstring, the closing `"""` should be on their own line):

```python
"""A multiline docstring.

With an extended description.
"""
```

In this case,  I would just keep it on one line by just saying:
| [...] taxonomy ID is not supported.

README.md
uniqueg, line 1: Don't forget to update the PR title and description (replace `org_id` with `tax_id`)

htsinfer/get_library_source.py
```diff
@@ -50,6 +52,7 @@ class GetLibSource:
         min_freq_ratio: Minimum frequency ratio between the first and second
             most frequent source in order for the former to be considered the
             library's source.
+        tax_id: Taxonomy ID of the organism.
```
uniqueg: Better: "tax_id: Taxonomy ID of the _sample source_."

htsinfer/get_library_source.py
uniqueg: Oh, my bad. Don't know what I saw.

htsinfer/models.py
```diff
@@ -356,6 +356,7 @@ class Args(BaseModel):
         records: Number of input file records to process; set to `0` to
             process all records.
         threads: Number of threads to run STAR with.
+        tax_id: Taxonomy ID of the source organism.
```
uniqueg: As above...

uniqueg (dismissed):
Okay, getting there. I already approve, because I don't think it will need another review - but please address those comments ;-)

README.md
balajtimate: Updated it

htsinfer/get_library_source.py
balajtimate: Refactored it

htsinfer/exceptions.py
balajtimate: Thanks for the review, I refactored it :)

uniqueg (approved):

PR #150, opened 2023-11-15
fix: add processed sample paths to mapping

### LINKED ISSUE #149: bug: STAR using uncompressed sample files when mate relationship is not
**Describe the bug**
With the revised mapping logic from #131, when the input samples are _not_ mates and the mate relationship needs to be determined by mapping the reads against the transcripts, STAR uses the original input sample files instead of the uncompressed subset of the samples, resulting in an error, as the STAR command is currently not built to handle `.gz` files. 

**To Reproduce**
Run HTSinfer in paired-end mode on files from different mates, one of them compressed, e.g.
```
htsinfer tests/files/first_mate.fastq tests/files/mixed_mates_compressed.fastq.gz
```

**Expected behavior**
Add the paths to the subset of samples as argument when calling to the mapping function in `get_library_type.py`

balajtimate:
### Description

Add the paths to the processed subset of samples to the mapping function in `get_library_type.py`, so that STAR makes use of them instead of the original (unprocessed) samples, when aligning to determinate the mate relationship. 

Fixes #149 

### Type of change

### Checklist

_or do not apply_.

      warnings
      works
      by the proposed changes

together.

### DIFF
--- a/htsinfer/get_library_type.py
+++ b/htsinfer/get_library_type.py
@@ -131,6 +131,7 @@ plit_mates: 'split_mates'>)
             self.mapping.library_type.relationship \
                 = StatesTypeRelationship.not_available
             self.mapping.library_source = self.library_source
+            self.mapping.paths = (self.path_1, self.path_2)
             self.mapping.evaluate()
             self._align_mates()
 
--- a/htsinfer/get_library_type.py
+++ b/htsinfer/get_library_type.py
@@ -131,7 +131,7 @@ plit_mates: 'split_mates'>)
             self.mapping.library_type.relationship \
                 = StatesTypeRelationship.not_available
             self.mapping.library_source = self.library_source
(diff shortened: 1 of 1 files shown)


uniqueg:
Hmm, coverage test is failing. Maybe you need to update the unit tests (or add one)?

uniqueg (changes requested):
Please check the unit tests

balajtimate:
I'm not sure why, [the detailed coverage run](https://app.codecov.io/gh/zavolanlab/htsinfer/commit/4c3396804edf2684bf063748ce2edd16999be751) says 100% (also for the PR). Can it be rerun?

uniqueg:
Really weird. And no, I don't know how to rerun it without rerunning all the tests. Well, I think I'll just merge it then. If it persists, we can still investigate

PR #157, opened 2024-01-09
fix: infer library type relationship

### LINKED ISSUE #153: feat: only map when lib source is given, argument to force mapping otherwise
Currently, when inferring the orientation of the reads, the sample reads are mapped against the whole (unfiltered) transcripts library, when the library source is either not inferred or not given as input. This alters the result of the orientation inference. 

The program should check whether the library source is inferred or given, and only subset the transcripts and map to those based on the source organism. Also implement a CLI argument to force the mapping, even when those conditions are not met (but also make it known that the results are based on all the transcripts and no filtering was done). 

**Additional context**
Should the same apply when inferring the library layout relationship by mapping? As in only map when lib source is known.

balajtimate:
### Description

- Add check to assign `first_mate` and `second_mate` when it can correctly be inferred from seq_id's (when seq_ids are not empty)
- Add check to only align when `library source` is either inferred or given by `--tax-id` argument (in case of `library type` as well as `read orientation` inference)
- Add `--outSAMorder PairedKeepInputOrder` to `mapping.py` to keep input order of reads when running STAR with multiple threads
- Update tests

In the case of paired-end libraries, the logic should be:

1. For the library type inference, check `seq_ids` of first and second pair, to decide if they fit the Casava format, and the lib type and relationship can be determined
2. If not, as in most SRA samples, align them separately to reference transcripts, but only if `library_source` is known
3. Compare alignments and decide relationship
4. For the orientation inference, based on the library type relationship determined:
    1. If `not_mates` or `not_available`, infer the orientation from the separately aligned results
    2. If  `split_mates`, align the samples in paired-end mode, but only if `library_source` is known

Fixes #153 

### Type of change

### Checklist

_or do not apply_.

      warnings
      works
      by the proposed changes

together.

### DIFF
--- a/htsinfer/get_library_type.py
+++ b/htsinfer/get_library_type.py
@@ -144,6 +144,8 @@ plit_mates: 'split_mates'>)
         samfile1 = pysam.AlignmentFile(str(alignment_1), 'r')
         samfile2 = pysam.AlignmentFile(str(alignment_2), 'r')
 
+        seq_id1 = None
+        seq_id2 = None
         previous_seq_id1 = None
         previous_seq_id2 = None
 
@@ -154,43 +156,52 @@ plit_mates: 'split_mates'>)
         concordant = 0
 
         for read1 in samfile1:
-            seq_id1 = read1.query_name
-            if seq_id1 != previous_seq_id1 \
-                    and previous_seq_id1 is not None:
-                mate1.append(reads1.copy())
-                reads1.clear()
-            if read1.reference_end:
-                reads1.append(read1)
+            if not read1.flag & (1 << 2) and isinstance(read1.query_name, str):
+                seq_id1 = read1.query_name
+                if (
+                    seq_id1 != previous_seq_id1 and
+                    previous_seq_id1 is not None
+                ):
+                    mate1.append(reads1.copy())
+                    reads1.clear()
+                if read1.reference_end:
+                    reads1.append(read1)
             previous_seq_id1 = seq_id1
         mate1.append(reads1.copy())
 
         read_counter = 0
         for read2 in samfile2:
-            seq_id2 = read2.query_name
-            if seq_id2 != previous_seq_id2 \
-                    and previous_seq_id2 is not None:
-                if self._compare_alignments(mate1[read_counter], reads2):
-                    concordant += 1
-                reads2.clear()
-                read_counter += 1
-            if read2.reference_end:
-                reads2.append(read2)
+            if not read2.flag & (1 << 2) and isinstance(read2.query_name, str):
+                seq_id2 = read2.query_name
+                if seq_id2 != previous_seq_id2 \
+                        and previous_seq_id2 is not None:
+                    if self._compare_alignments(mate1[read_counter], reads2):
+                        concordant += 1
+                    reads2.clear()
+                    read_counter += 1
+                if read2.reference_end:
+                    reads2.append(read2)
             previous_seq_id2 = seq_id2
 
         if self._compare_alignments(mate1[read_counter], reads2):
             concordant += 1
 
-        if (concordant / read_counter) >= self.cutoff:
-            self.results.relationship = (
-                StatesTypeRelationship.split_mates
-            )
-            self.mapping.library_type.relationship \
-                = StatesTypeRelationship.split_mates
-            self.mapping.mapped = False
-            self.mapping.star_dirs = []
+        if read_counter > 0:
+            if (concordant / read_counter) >= self.cutoff:
+                self.results.relationship = (
+                    StatesTypeRelationship.split_mates
+                )
+                self.mapping.library_type.relationship \
+                    = StatesTypeRelationship.split_mates
+                self.mapping.mapped = False
+                self.mapping.star_dirs = []
+            else:
+                self.results.relationship = (
+                    StatesTypeRelationship.not_mates
+                )
         else:
             self.results.relationship = (
-                StatesTypeRelationship.not_mates
+                StatesTypeRelationship.not_available
             )
 
         samfile1.close()
--- a/htsinfer/get_library_type.py
+++ b/htsinfer/get_library_type.py
@@ -127,13 +127,19 @@ plit_mates: 'split_mates'>)
                 self.mapping.library_type.relationship = (
                     StatesTypeRelationship.split_mates
                 )
-        else:
+        elif self.library_source.file_1.short_name is not None \
+                and self.library_source.file_2.short_name is not None:
             self.mapping.library_type.relationship \
                 = StatesTypeRelationship.not_available
             self.mapping.library_source = self.library_source
             self.mapping.paths = self.path_1, self.path_2
             self.mapping.evaluate()
             self._align_mates()
+        else:
+            LOGGER.warning(
+                "Library source is not determined, "
+                "mate relationship cannot be inferred by alignment."
+            )
 
     def _align_mates(self):
         """Decide mate relationship by alignment."""
--- a/htsinfer/get_library_type.py
+++ b/htsinfer/get_library_type.py
@@ -113,7 +113,7 @@ plit_mates: 'split_mates'>)
             ids_2: As `ids_1` but for the putative second mate file.
         """
         self.results.relationship = StatesTypeRelationship.not_mates
-        if ids_1 == ids_2:
+        if ids_1 == ids_2 and len(ids_1) > 0 and len(ids_2) > 0:
             if (
                 self.results.file_1 == StatesType.first_mate and
                 self.results.file_2 == StatesType.second_mate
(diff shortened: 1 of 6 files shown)


htsinfer/get_library_type.py
```diff
@@ -113,7 +113,7 @@ def _evaluate_mate_relationship(
             ids_2: As `ids_1` but for the putative second mate file.
         """
         self.results.relationship = StatesTypeRelationship.not_mates
-        if ids_1 == ids_2:
+        if ids_1 == ids_2 and len(ids_1) > 0 and len(ids_2) > 0:
```
uniqueg: I guess you can just write `if ids_1 == ids_2 and ids_1 and ids_2:`, as a logical check will return `False` if a list is empty and `True` only if it has at least one item. And to make it a bit more efficient and easier to digest, I would change the order like so:

```py
if ids_1 and ids_2 and ids_1 == ids_2:
```

htsinfer/get_library_type.py
```diff
@@ -127,13 +127,22 @@ def _evaluate_mate_relationship(
...
             self.mapping.library_source = self.library_source
             self.mapping.paths = self.path_1, self.path_2
             self.mapping.evaluate()
             self._align_mates()
+        else:
+            self.results.relationship = StatesTypeRelationship.not_available
+            LOGGER.debug(
+                "Library source is not determined, "
+                "mate relationship cannot be inferred by alignment."
+            )
```
uniqueg: I think this should rather be a warning? Or does this lead to a warning being raised elsewhere? Anyway, we wanna be consistent, so if other failed "inferences" don't lead to a warning, then this one shouldn't either.

Also, I think the message should include both of the other cases, not just the inability to do alignments. Or just keep it simple and simply say that the mate relationship could not be determined.

htsinfer/get_library_type.py
```diff
@@ -154,47 +165,68 @@ def _align_mates(self):
...
 
         for read1 in samfile1:
-            seq_id1 = read1.query_name
-            if seq_id1 != previous_seq_id1 \
-                    and previous_seq_id1 is not None:
-                mate1.append(reads1.copy())
-                reads1.clear()
-            if read1.reference_end:
-                reads1.append(read1)
+            if not read1.flag & (1 << 2) and isinstance(read1.query_name, str):
```
uniqueg: I think this is one of the few cases where I think a comment in the code base would be helpful to read the code. Bitwise operations aren't so common and the multiple conditions further make it a bit hard to follow this.

htsinfer/get_library_type.py
```diff
@@ -154,47 +165,68 @@ def _align_mates(self):
...
+                    ):
+                        concordant += 1
+                    reads2.clear()
+                    read_counter += 1
+                if read2.reference_end:
+                    reads2.append(read2)
             previous_seq_id2 = seq_id2
 
-        if self._compare_alignments(mate1[read_counter], reads2):
+        if read_counter < len(mate1) and self._compare_alignments(
```
uniqueg: Why does the number of aligned reads in file 2 need to be smaller than those of file 1 here? I find the code a bit difficult to follow (not your fault, I know) :-S

htsinfer/get_library_type.py
```diff
@@ -154,47 +165,68 @@ def _align_mates(self):
...
-            )
-
-        samfile1.close()
-        samfile2.close()
+            if ratio >= self.cutoff:
+                self.results.relationship = (
+                    StatesTypeRelationship.split_mates
+                )
+                self.mapping.library_type.relationship \
+                    = StatesTypeRelationship.split_mates
```
uniqueg: I would use the `()` way of wrapping consistently (like in the line above)

uniqueg: By the way, this is possible in Python (although here it will probably lead to a line that is too long):

```py
a = b = some_expression
```

i.e., you can assign the same value to multiple variables in the same statement.

htsinfer/get_read_orientation.py
```diff
@@ -75,8 +75,16 @@ def evaluate(self) -> ResultsOrientation:
...
+        if not self.mapping.mapped and (
+            self.library_source.file_1.short_name is not None or
+            self.library_source.file_2.short_name is not None
+        ):
             self.mapping.evaluate()
+        else:
+            LOGGER.debug(
+                "Library source is not determined, "
+                "read orientation cannot be inferred by alignment."
+            )
```
uniqueg: As above: Please check how other inference failures are logged and use the same logging level. If it's always `debug` then ignore this comment.

uniqueg (changes requested):
This looks good to me (minus the very few minor comments), but I have to be honest in that it is a bit hard to follow the code in the library type model since the introduction of the mappings. I am pretty confident given the extensive tests, but I propose that you play this through with some small real world examples as well to check if this really behaves as desired from end to end.

htsinfer/get_library_type.py
balajtimate: Thanks, I changed that in the commit.

htsinfer/get_library_type.py
balajtimate: I thought so too at first, but all the information regarding the process (the seq id format, if alignment is needed etc) are all on the debug level, so getting a warning about the mate relationship determination when running it on info level seems a bit weird, doesn't it?
But yeah, I update the message to say the relationship cannot be determined.

htsinfer/get_library_type.py
balajtimate: Added comment for clarification.

htsinfer/get_library_type.py
balajtimate: Right, fixed it.

htsinfer/get_library_type.py
balajtimate: Refactored it in newest commit.

balajtimate:
To summarize: refactored `_align_mates` in `get_library_type.py` to correctly determine the number of aligned reads as well as the number of concordant reads. 

When iterating through the reads, the problem during `_compare_alignments` was that the order of the reference names were scrambled. Turns out, this was the result of STAR: it outputs the aligned reads in the same order as the input read, _except_ when running it on multiple threads, which I have been doing since day 1. When I reran the samples on the default 1 thread, the output of the alignment matched between the two samples, and the concordant reads were calculated correctly. So one solution was to sort reads by ref name using `pysam`, which outputs an extra sorted BAM file, but ultimately updating the STAR command with `--outSAMorder PairedKeepInputOrder` seems to solve this, so there is no extra sorting needed, the output will be in the same order as input (it's also possible to sort the aligned reads with STAR, but only according to coordinates, and I think it'd be better to have sorted according to ref names. Btw, is there any advantage in this case for using SAMs instead of BAMs? I think `mapping.py` could be updated to output BAMs, maybe in a separate PR). 

For checking the ratio between the concordant and the aligned reads (for `_update_relationship`), I choose the lowest of the two alignments. Also, I'll run it on more samples and check the ratio, but I think we could lower it a bit, like 90% (instead of 95% default now) for the ratio / aligned reads to be considered paired end library.

htsinfer/mapping.py
```diff
@@ -299,6 +299,7 @@ def build_star_command(
                 "--runThreadN", f"{str(self.threads_star)}",
                 "--genomeDir", f"{str(index_dir)}",
                 "--outFilterMultimapNmax", "50",
+                "--outSAMorder", "PairedKeepInputOrder",
```
uniqueg, line 302: But does this handle cases where the inputs, as provided by the users, are unsorted/weirdly sorted?

If not, I guess it's fine to keep it like this for now, but maybe you could document the assumption of properly sorted files for library type inference somewhere, and you could open a low priority issue for implementing sorting of the inputs so as to remove that assumption.

htsinfer/get_library_type.py
```diff
@@ -159,57 +158,66 @@ def _align_mates(self):
...
-                    mate1.append(reads1.copy())
-                    reads1.clear()
-                if read1.reference_end:
-                    reads1.append(read1)
+            seq_id1 = read1.query_name
+            if (
+                seq_id1 != previous_seq_id1 and
+                previous_seq_id1 is not None
+            ):
+                mate1.append(reads1.copy())
```
uniqueg, line 175: The solution still requires a lot of memory. It's fine but please open a low priority issue to improve this implementation at some point in the future.

uniqueg (dismissed):
Approving, but please take care of the two comments first :pray:

uniqueg:
Oh, and please update the PR description to reflect all of the latest changes. As far as I understood, you also _fixed_ some things that never really worked.

htsinfer/mapping.py
balajtimate: > But does this handle cases where the inputs, as provided by the users, are unsorted/weirdly sorted?

Hm, I guess not, but at least the order doesn't get messed up further by STAR. Added comments about the assumption of sorted order for input files, and opened #159

htsinfer/get_library_type.py
balajtimate: Opened #158

uniqueg (approved):

PR #161, opened 2024-01-23
feat: improve lib type inference for SRA PE samples

### LINKED ISSUE #160: feat: improve lib type relationship inference for SRA samples
With #157, the library type relationship is determined by aligning the reads separately and comparing the ratio of concordant pairs to the number of aligned reads. This works great for SRA samples where the `seq_ids` don't confirm to the Illumina seq_id formats. The problem is with samples where the `seq_ids` are in the appropriate format, but they don't contain any information regarding their mate, so HTSinfer treats them as `single`. In this case, it would be beneficial to double-check the relationship by alignment, trying to minimize the number of samples deemed "single", when two files were given as input.

Also, when the  the relationship is deemed `split_mates` based on the ratio of concordant reads, the type of library for the samples still remain either `single` (in the case of above example) or `null` (for SRA samples where the seq_id is not in the correct format). In this case, I would say it's safe to assume that input 1 is the `first_mate` and input 2 is `second_mate`, but since it's not _confirmed_ from the seq_ids, so additional categories could be added (`first_mate_assumed`, `second_mate_assumed`) as suggested, and the lib type updated after the alignment. 

Check the relationship by forcing the alignment when both inputs are deemed "single"; update lib type for individual samples based on alignment results.

balajtimate:
### Description

- Expand `_evaluate_mate_relationship` for library type relationship inference to align samples when two input files are given but both are deemed `single`
- Add additional library type states `first_mate_assumed` and `second_mate_assumed`, when the relationship is determined to be `split_mates`, refactor `_update_relationship_type` helper function to assign states types
- Update tests
- Lower `lib_type_mates_cutoff` to 0.85 based on results from #160 

Closes #160 

### Type of change

### Checklist

_or do not apply_.

      warnings
      works
      by the proposed changes

together.

### DIFF
--- a/htsinfer/cli.py
+++ b/htsinfer/cli.py
@@ -232,7 +232,7 @@ def parse_args() -> argparse.Namespace:
         dest="lib_type_mates_cutoff",
         metavar="FLOAT",
         type=float,
-        default=0.95,
+        default=0.85,
         help=(
             "minimum fraction of mates that can be mapped to compatible loci "
             "and are considered concordant pairs / all mates"
--- a/htsinfer/get_library_type.py
+++ b/htsinfer/get_library_type.py
@@ -127,6 +127,21 @@ plit_mates: 'split_mates'>)
                 self.mapping.library_type.relationship = (
                     StatesTypeRelationship.split_mates
                 )
+            # Infer mate relationship, even when assumed to be single
+            elif (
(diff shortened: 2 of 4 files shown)


uniqueg (approved):
LGTM :)

PR #163, opened 2024-02-05
build: create data package

balajtimate:
### Description

- Add `__init__.py` to `data` directory, to mark is as subpackage and include it in conda builds
- Update versioning in `htsinfer/__init__.py`

### Type of change

### Checklist

_or do not apply_.

      warnings
      works
      by the proposed changes

together.

### DIFF
--- a/htsinfer/__init__.py
+++ b/htsinfer/__init__.py
@@ -2,4 +2,4 @@
 
 from htsinfer.htsinfer import HtsInfer  # noqa:F401
 
-__version__ = "0.9.0"
+__version__ = "0.10.0"


balajtimate:
There's a temporary issue with coverage, but the unit tests run normally, the report just cannot be submitted.

uniqueg:
Yeah, I had the same. Annoying 😅

htsinfer/__init__.py
```diff
@@ -2,4 +2,4 @@
 
 from htsinfer.htsinfer import HtsInfer  # noqa:F401
 
-__version__ = "0.9.0"
+__version__ = "0.10.0"
```
uniqueg, line 5: Why is this from 9 to 10? I thought we were already at 10? :)

balajtimate: Yes, but it wasn't updated here, we would re-release `v0.10.0`

uniqueg (approved):

PR #168, opened 2024-10-02
refactor: update transcript list

### LINKED ISSUE #56: Improve library source inference
- Consistently update set of reference transcripts for library source inference and include more genes for better accuracy
- Include organisms/sources from all clades in Ensembl: Bacteria, Plants, Fungi, Metazoa etc.

### LINKED ISSUE #108: Analyze organism/source
1. Compare predictions with reported source
2. Quantify accuracy: correct, false, undecided

One goal is to minimize false predictions. Here we can try playing with `--library-source-min-match-percentag` and `--library-source-min-frequency-ratio`.

However, another goal is to also make sure that we get reliable predictions at least for the most common model organism (not useful that we never get a mouse because based on our rRNA gene transcripts we cannot sufficiently distinguish between mouse and shrew mouse). About this, there are basically two things we can do:
* Don't support so many organisms, and particularly those that are closely related
* Add transcripts of more genes, including ones with a more dynamic evolution

balajtimate:
### Description

- Update `transcripts.fasta.gz` to include transcript sequences from an updated set of small subunit RP genes (RPSA, RPS2, RPS3, RPS9, RPS11, RPS13, RPS14, RPS15, RPS15A, RPS16, RPS18, RPS20, RPS23, RPS29) that are consistent across the domains of Bacteria, Archaea and Eukaryotes (so in the future Bacteria can also be included). The RP genes were selected based on https://bangroup.ethz.ch/research/nomenclature-of-ribosomal-proteins.html. Human genes and their orthologs were downloaded from [Ensembl main (Vertebrates)](ensembl.org), [Ensembl Plants](plants.ensembl.org), [Ensembl Fungi](fungi.ensembl.org) and [Ensembl Metazoa](metazoa.ensembl.org). The set now consists of 14410 transcripts from 385 organisms.
- Update to correct current version `v0.11.0`
- Add disable statement to pylint `too-many-positional-arguments` in `get_read_layout.py` (related to changes in pylint [v3.3](https://pylint.readthedocs.io/en/latest/whatsnew/3/3.3/index.html#))

Closes #108
Closes #56

### Type of change

### Checklist

_or do not apply_.

      warnings
      works
      by the proposed changes

together.


balajtimate:
With the updated transcript list and the new set of SRA samples (which _does_ contain 720 samples), we have the best result for library source inference yet:


I think with this we can close #108 and #56 (improving lib source inference with a consistent set of RP genes)

balajtimate:
Also, the checks are failing because there's an ongoing [issue](https://github.com/conda-incubator/setup-miniconda/issues/366) with `setup-miniconda` as a result of mamba v2 release. There seems to be a fix on the way, should we wait for it or pin mamba to 1.5.10 @uniqueg? That worked in the previous [commit](https://github.com/zavolanlab/htsinfer/pull/168/commits/17fb60004ba90f91169f676add61aa04703fd0e8).

uniqueg (approved):

uniqueg:
If you think a fix will come within the next few days, then wait - otherwise pin.

PR #169, opened 2024-10-08
feat: output read orientation fractions to json

balajtimate:
### Description

- Add json output of the fractions of read orientation (number of mapped reads, fraction of SF, SR and final orientation) per input sample
- Clean up debug logging

### Type of change

### Checklist

_or do not apply_.

      warnings
      works
      by the proposed changes

together.


htsinfer/get_read_orientation.py
```diff
@@ -176,13 +178,31 @@ def process_single(
...
-            f"Required number of mapped reads pairs: {self.min_mapped_reads}"
+            f"Required number of mapped reads: {self.min_mapped_reads}"
         )
-        LOGGER.debug(f"Number of reads mapped: {reads}")
-        LOGGER.debug(f"Fraction of states: {fractions_all_states}")
-        LOGGER.debug(f"Orientation: {orientation}")
+        LOGGER.debug(f"Number of mapped reads: {orient_df.iloc[0, 0]}")
+        LOGGER.debug(f"Fraction of SF: {orient_df.iloc[0, 1]}")
+        LOGGER.debug(f"Fraction of SR: {orient_df.iloc[0, 2]}")
+        LOGGER.debug(f"Orientation: {orient_df.iloc[0, 3]}")
```
uniqueg, line 191: I still think these kinda messages aren't really debug level, they are very informative for the user (despite being in the output), so I think `INFO` would be much better suited here. If people would like to see only warnings, they can change the verbosity accordingly.

That being said, if the same principle is followed for other functionalities, i.e., that results/info messages are treated as debug it's not a matter for _this_ PR, but a separate one, that changes log levels for these kinda messages across all functionialities.

However, if it's only this one, please change here.

htsinfer/get_read_orientation.py
```diff
@@ -176,13 +178,31 @@ def process_single(
...
+            ),
+            'Fraction SR': fractions[0].get(
+                StatesOrientation.stranded_reverse
+            ),
+            'Orientation': orientation.value
+        }])
+
+        orient_df = self.create_orient_df(
+            reads, fractions_all_states, orientation, paired=False
+        )
```
uniqueg: Why not call `create_orient_df()` here, since you already defined it (like you do for paired reads)?

htsinfer/get_read_orientation.py
```diff
@@ -338,3 +377,77 @@ def sum_dicts(*dicts: Dict[Any, float]) -> Dict[Any, float]:
             for key, num in dct.items():
                 result[key] += num
         return dict(result)
+
+    def create_orient_df(  # pylint: disable=R0917
+            self,
```
uniqueg: Do you really need `self` here? Seems like you don't, so I guess you should make it a `@staticmethod`.

uniqueg (changes requested):
Some minor issues. In addition to the ones mentioned in line, I'm wondering if there aren't some docs that need to be updated? Example outputs maybe?

htsinfer/get_read_orientation.py
balajtimate: Yes, they're all `DEBUG` for all other modules. I'll change them in another PR then.

htsinfer/get_read_orientation.py
balajtimate: Oops, I missed that, fixed that in the commit

balajtimate:
No docs update needed right now. I'll rewrite the full documentation on RtD, there it will be included.

uniqueg:
> No docs update needed right now. I'll rewrite the full documentation on RtD, there it will be included.

Oh, never mind. Had only a brief look and thought that you'd ended up including the stats in the main output after all. But seeing now that you're just writing it into a separate file, as discussed 👍

uniqueg (approved):

PR #170, opened 2024-10-24
perf: use estimated counts instead of TPM

balajtimate:
### Description

- Use 'estimated counts' of transcript abundance from Kallisto (instead of TPM) for library source inference
- Update `transcripts.fasta.gz` to include additional organisms and transcripts from orthologues, which were previously removed (now it includes transcripts from **387** organisms)
- Repetitive sequences in transcripts have been masked using [RepeatMasker](https://www.repeatmasker.org/)

### Type of change

### Checklist

_or do not apply_.

      warnings
      works
      by the proposed changes

together.


htsinfer/get_library_source.py
```diff
@@ -254,10 +254,10 @@ def get_source_expression(
 
         Returns:
             Data frame with columns `source_ids` (a tuple of source short name
-                and taxon identifier, e.g., `("hsapiens", 9606)`) and `tpm`,
-                signifying the percentages of total expression per read source.
-                The data frame is sorted by total expression in descending
-                order.
+                and taxon identifier, e.g., `("hsapiens", 9606)`) and
+                `est_counts`, signifying the percentages of total expression
+                per read source. The data frame is sorted by total expression
```
uniqueg: I'm not sure it is accurate to speak of "expression" when we use counts. Perhaps best to refer to "total estimated counts".

uniqueg (changes requested):

htsinfer/get_library_source.py
balajtimate: Thanks, addressed it in commit

uniqueg (approved):

PR #171, opened 2024-10-29
refactor: update logging level

balajtimate:
### Description

- Change logging levels from `DEBUG` to `INFO` to clarify when background processes (e.g. alignment with STAR or kallisto quantification) are running, as well as give extra info (number of aligned reads per sample, fraction of orientations)

### Checklist

_or do not apply_.

      warnings
      works
      by the proposed changes

together.

### DIFF
--- a/htsinfer/get_library_stats.py
+++ b/htsinfer/get_library_stats.py
@@ -88,7 +88,8 @@ class GetLibStats:
             FileProblem: Could not process FASTQ file.
         """
         LOGGER.debug(
-            f"Extracting read length statistics in: {fastq}")
+            f"Extracting read length statistics in: {fastq}"
+        )
         min_len: int = 1000000
         max_len: int = 0
         total_lengths: int = 0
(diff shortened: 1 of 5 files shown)


htsinfer/get_library_stats.py
```diff
@@ -88,7 +88,8 @@ def fastq_get_stats_read_length(fastq: Path) -> Tuple[
             FileProblem: Could not process FASTQ file.
         """
         LOGGER.debug(
```
uniqueg, line 90: Isn't this sort of the same level/granularity of information that you now log via `info` instead of `debug`? :)

uniqueg (changes requested):

htsinfer/get_library_stats.py
balajtimate: Hm, there's actually already a log message in line 54, this is sort of redundant. Wouldn't it make sense to keep it `debug` for specifically the `fastq_get_stats_read_length` function?

uniqueg (approved):

